PR #4, opened 2019-12-13
Add validation

ylt:
Use `"@openactive/data-model-validator` to validate that the responses are structurally correct.


ylt:
Example:
image

packages/openactive-integration-tests/test/flows/booking/book.js
```diff
@@ -1,5 +1,6 @@
 const chakram = require("chakram");
 const RequestHelper = require("../../helpers/request-helper");
+const validationHelper = require("../../helpers/validation-helper");
```
henryaddison: maybe call directory shared-tests to make it clearer it defines tests rather than being a standalone little helper function

henryaddison: then file could just be validation and the const could be sharedValidationTests (or something)

packages/openactive-integration-tests/test/helpers/validation-helper.js
```diff
@@ -0,0 +1,49 @@
+const assert = require("assert");
+const { validate: validateData } = require("@openactive/data-model-validator");
+
+// we use a getter method as we aren't able to get the test scope from here
+function validate(getter, name, options = {}) {
```
henryaddison: then this function could be called something like `shouldBeValidResponse`

packages/openactive-integration-tests/test/flows/booking/book.js
ylt: fine with it being called just /shared?

PR #7, opened 2020-01-06
Improve usability

ylt:
- Make npm install also install deps of subpackages
- Add a global setup to check that the broker is available
- Make the broker logging more useful + less confusing


ylt:
@thtg88 oh sorry, should've better explained it. :)

I am currently writing the documentation / readme. This PR is basically to remove unnecessary steps/details from the readme to keep it simple for the user.

Issue #10, opened 2020-01-22
Remove duplication between book-shared.js files as far as possible


nickevansuk:
Lots of duplicate code between "book-shared.js”, can we remove this duplication as far as possible? I.e. rationalise code that defines each flow and scenario to ensure maximum code reuse

ylt:
> I.e. rationalise code that defines each flow and scenario to ensure maximum code reuse
Flows have been deduplicated in https://github.com/openactive/openactive-test-suite/pull/30

An idea is that we may be possible to further remove duplication by using shared examples. We can define the various expectations elsewhere, and include those into each test suite that require them. (i.e. we're often checking C1 responses, so can have a set of common tests for those)

PR #30, opened 2020-02-17
Reduce duplication

ylt:
Moves the chaining and variable assignment / passing logic out of the individual test suites into a couple of helper classes.

- Request-State-Helper - helper object, implements most of the methods of the RequestHelper, however it keeps state on itself which it passes to the various requests, and stores extracted values from responses against itself.
- FlowHelper - it's only job is to run the various tasks in sequence. It wraps a provided Request-State-Helper

Addresses #10

### DIFF
--- a/packages/openactive-integration-tests/test/helpers/flow-helper.js
+++ b/packages/openactive-integration-tests/test/helpers/flow-helper.js
@@ -1,0 +1,55 @@
+const RequestHelper = require("./request-helper");
+const pMemoize = require("p-memoize");
+
+class DefaultFlowHelper {
+  constructor (helper) {
+    this.helper = helper;
+  }
+
+  log (msg) {
+    this.helper.logger.log(msg);
+  }
+}
+
+class FlowHelper extends DefaultFlowHelper {
+  getMatch = pMemoize(async getMatch => {
+    return this.helper.getMatch();
+  });
+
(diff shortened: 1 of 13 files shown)


packages/openactive-integration-tests/test/helpers/flow-helper.js
```diff
@@ -0,0 +1,55 @@
...
+    await this.U();
+
+    return await this.getOrderPromise;
+  });
+}
+
+module.exports = {
+  DefaultFlowHelper,
+  FlowHelper
+};
```
henryaddison: What is DefaultFlowHelper for? It appears to only be used as a base class for FlowHelper?

packages/openactive-integration-tests/test/helpers/request-state-helper.js
```diff
@@ -0,0 +1,118 @@
...
+  // var uuid;
+  // var totalPaymentDue;
+  // var orderItemId;
+  //
+  // var c1Response;
+  // var c2Response;
+  // var bResponse;
+  // var uResponse;
+  // var getOrderPromise;
+  // var ordersFeedUpdate;
```
henryaddison: I think it may be worth keeping a list of the state set but not as a commented out list of variable declarations

PR #31, opened 2020-02-18
Changes to pass validation

nickevansuk:
@ylt so the live reference implementation now passes validation against the latest `data-models` using the `data-model-validator`. I had to fix some minor bugs in this test suite to get them to pass (see this PR).

Rather than attempting to merge these in to your work, as I know you're doing some refactoring here, I've left this as a PR just so you can see the fixes.

Would you be able to just port these over to your new refactored code? You should see that all validation tests now pass.

Please do check out the results of C1, as it looks like the validation isn't running properly for that for some reason, but you may have already fixed this in your refactor?

### DIFF
--- a/packages/openactive-broker-microservice/app.js
+++ b/packages/openactive-broker-microservice/app.js
@@ -425,7 +425,7 @@ function monitorPage(rpde, pageNumber) {
     }
   });
 
-  setTimeout(x => getRPDE(rpde.next, monitorPage), 5000);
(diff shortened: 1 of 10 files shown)


ylt:
Github is showing the merged in changes a little weirdly, looking into it.

ylt:
Solved - changing the base away from master and back again fixed it. So, just Github bug.

Issue #26, opened 2020-01-27
 Output a test result certification zip file


nickevansuk:
The integration tests should include a script entry in the package.json to output a single Test Result Certification Zip File, which is a zip file containing all test results in a well-defined structure. This script must use only npm dependencies (e.g. gulp, and gulp-zip), and not rely on the local operating system to create the zip, to maximise compatibility.

In addition to containing the console output and log files (as evidence of the requests and responses from the implementation under test), the zip file should contain a machine-readable version of the results, to allow a separate service to easily verify that all tests pass.

thill-odi:
Deferred.

Issue #18, opened 2020-01-22
Clearly reference the relevant log file


nickevansuk:
Ensure the errors returned by Jest clearly reference the relevant log file where the request/response content can be found, to make it easier to navigate the output

Issue #15, opened 2020-01-22
Include other test failures in log files


nickevansuk:
Include non-validation test failures in log file, so they are all in one place

ylt:
This is a little tricky to integrate with. Logging will need to refactored to be handled at a slightly more global level, however there are some new complexities that need figuring out.

https://jestjs.io/docs/en/configuration#globalsetup-string
> "Note: Any global variables that are defined through globalSetup can only be read in globalTeardown. You cannot retrieve globals defined here in your test suites."

So, we may very well need multiple loggers that write to the same files. This should be possible to safely do.

Approach I'm going with is to use a reporter. The reporter seems to only be invoked once the tests are done as a whole, so shouldn't have any problems with multiple writes at the same time.

---

Little investigation into the ordering of things.. (looking into whether to use a global teardown in combination with the reporter)

![openactive-integration-tests (zsh) 2020-02-19 10-25-35](https://user-images.githubusercontent.com/705668/74827354-33f72c00-5305-11ea-8aa3-f7fd0355e532.png)

("global teardown" is a script defined as the [globalTeardown](https://jestjs.io/docs/en/configuration#globalteardown-string), with the rest of the messages coming from a[ custom reporter](https://jestjs.io/docs/en/configuration#reporters-arraymodulename--modulename-options) defining the "onRunStart, onTestStart, onTestResult, onRunComplete" methods)

nickevansuk:
As discussed on call just now - we should expand this issue to include making the presentation of the log files as useful as possible, so that they can be used instead of all the (fairly confusing) console output

ylt:
The console output can be fully changed as well, especially with the changes I need to make.

ylt:
At the moment, it's using the `default` reporter (the one that's showing the stack trace, etc), in combination with the `spec` reporter (the checklist).

Current plan is to make the tests output to a more predictable file path.

On completion of a test suite, have the reporter pick up that file and output a more pretty file that also includes the test status. At that point it's just a case of also console.log-ing it out. Perhaps that output is good enough then to completely disable the current reporters.

nickevansuk:
Just to note on this one - if there are timeout issues it's not currently very obvious what's going on as stuff just starts randomly failing (C1 responses don't show up, the validator doesn't get triggered, etc)

Looking at the .json output there are messages like:
[...]

Issue #13, opened 2020-01-22
Include a description of expected behaviour in tests


nickevansuk:
Include expected behaviour in the tests somehow, to make it clear what’s happening. Perhaps a description can be added, e.g. "Opportunity.startDate is in the past, which means it is not possible to cancel"

ylt:
It's not entirely clear what the best way of doing these is. It could be done using the describe blocks, but having longer text looks rather unfriendly in the console.

PR #33, opened 2020-02-20
Better logging

ylt:
Implements:
 - https://github.com/openactive/openactive-test-suite/issues/15 - "Include other test failures in log files"
 - https://github.com/openactive/openactive-test-suite/issues/13 - "Include a description of expected behaviour in tests"
 - https://github.com/openactive/openactive-test-suite/issues/18 - "Clearly reference the relevant log file"
 - Groundwork necessary for https://github.com/openactive/openactive-test-suite/issues/26 - "Output a test result certification zip file" (now generates machine readable data, however no zipping)


ylt:
Almost there, main thing now is that output of specs is duplicated.

openactive-integration-tests (zsh) 2020-02-20 12-08-18

ylt:
@nickevansuk Any suggestions on how this should look? This is the console version of the output:

image
image

nickevansuk:
Nice! Is this the only output? Or is do the outputted reports also still include the actual requests and response JSON, which are useful for debugging with the list of tests that are running

ylt:
The plan is to generate both, just a slightly different template file for each (so console can exclude the json, but files can include it).

nickevansuk:
Excellent! Ye output looks great! The usefulness will obviously come down to how the individual tests are named, which works well, so e.g. "should return newly created event" will need to be self-documenting in terms of what that means or is checking for. I guess that's about them having a good balance between being concise and being descriptive.

Only minor comment would be could we change the exclamation mark emoji for failure to an "X" (as in https://github.com/openactive/openactive-test-suite/issues/15#issuecomment-588159255), to make it clearer it's a failure?

nickevansuk:
Was there anything else you were considering here? Seems useful from the screenshots?

ylt:
Sure, will change the emoji. Almost have the file output there, will show that as well once ready.

nickevansuk:
Amazing @ylt !

nickevansuk:
The only other thing that might be worth including in the console is something that indicates that requests and responses were actually sent.

I've noticed a few issues with the tests where (for whatever reason) the request for C1 doesn't actually get made (e.g. due to timeout, async issues, because the endpoint doesn't exist because the implementation is still in progress, etc)

For this would it be possible to:
a) Just return a test failure that says the Call to C1 endpoint failed
b) Or perhaps better return a separate line in the console for C1 request sent, C1 response received

These might be achieved with some additional tests that are run first within "C1"

Helps to debug situations where either (i) the endpoints just aren't working at all (e.g. URL is wrong, proxy issues, etc) or (ii) the test suite isn't actually making the requests

ylt:
Made it generate a markdown file, an example: https://gist.github.com/ylt/82a2d12d5ab94c1119e233800a5ea378

nickevansuk:
Love it! The great thing about the markdown is that viewing it in a text editor works too, and something like VS Code can render it easily.

Can we include the status code of responses in the output (thinking about how I might go ahead and debug the markdown file example)? Especially when there's a blank in both request and response (?) - I guess see previous comment https://github.com/openactive/openactive-test-suite/pull/33#issuecomment-589029502

Also not sure the correct validation mode is being used, or maybe the latest `data-models` and `data-model-validator` is not being used? As the validation errors don't seem correct?

ylt:
Yeah, it is supposed to be displaying status code there. I'm currently looking into it. 

It's probably an older version of master. I'll try updating to latest master and check the versions.

```handlebars
# {{{ title }}}

{{{ description }}}

Flows:
{{#each flow}}
## {{{ @key }}}
### Request
```json
{{{ json request }}}
```
### Response
{{{ response.status }}}
```json
{{{ json response.body }}}
```
### Validations
{{#each response.validations}}
* {{{ validationIcon severity }}} {{{firstLine message}}}
{{/each}}
### Specs
{{#each response.specs}}
* {{{ specIcon status }}} {{{firstLine title}}}
{{/each}}

{{/each}}
```

ylt:
Did try pasting in the handlebars template there, but Github didn't like markdown codeblocks inside a code block.
https://github.com/openactive/openactive-test-suite/blob/118dcd150339480f25054978752ea3fa328f046b/packages/openactive-integration-tests/test/report-templates/report.md.handlebars

nickevansuk:
Sure. Perhaps it should be prefixed with "Response Status Code: {{{ response.status }}}" (assuming that response.status is just e.g. '200')?

ylt:
> It would be useful to have some documentation telling me about the existence of the file output as well as the console output when after I've run the tests and to explain the different files
It does actually do this :)

openactive-integration-tests (zsh) 2020-02-21 14-05-31

henryaddison:
Yes but that gets a bit lost in the volume of output, particularly as the README doesn't even let me know there is extra output.

nickevansuk:
Also there's an issue where the code on this branch doesn't seem to record the response from B in the logs for some reason

ylt:
Picked up one of the code review comments as a new PR: https://github.com/openactive/openactive-test-suite/pull/35
> In JSON files in output/meta, might it be better to use an array for the flow to communicate order? Also I don't know but might a step be repeated?

Not entirely sure if it's appropriate to finish for now, but at least good progress towards it.

Issue #32, opened 2020-02-18
Validation mode should recognise errors


nickevansuk:
`validationMode` of `OpenBookingError` should be used for responses with a status code that is not `200`, from C1, C2 and B in all tests

PR #36, opened 2020-02-25
Add validation error handling

ylt:
Implements #32  

Makes the validation deterministically use a "OpenBookingError" if the response code is a non 2xx. Also adds in a custom validation error message in the case that the body is empty (and thus nothing to validate).

```
• C2
    ❌️ Server returned an error 500 (Internal Server Error) with an empty body.
    ❌️ should return 200 on success
    ❌️ offer should have price of 14.95
```

### DIFF
--- a/packages/openactive-integration-tests/test/flows/book-and-cancel/book-and-customer-cancel-success/book-shared.js
+++ b/packages/openactive-integration-tests/test/flows/book-and-cancel/book-and-customer-cancel-success/book-shared.js
@@ -66,7 +66,7 @@ function performTests(dataItem) {
       });
     });
 
-    sharedValidationTests.shouldBeValidResponse(() => state.c1Response.body, "C1", logger, {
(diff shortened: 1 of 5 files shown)


packages/openactive-integration-tests/test/flows/book-and-cancel/book-and-customer-cancel-success/book-shared.js
```diff
@@ -180,7 +180,7 @@ function performTests(dataItem) {
       );
     });
 
-    sharedValidationTests.shouldBeValidResponse(() => state.ordersFeedUpdate.body.data, "Orders feed", logger, {
+    sharedValidationTests.shouldBeValidResponse(() => state.ordersFeedUpdate.body, "Orders feed", logger, {
```
thtg88: any reason why you are not using `state.ordersFeedUpdate` instead of `state.ordersFeedUpdate.body`?

Issue #34, opened 2020-02-21
Multiple improvements to test suite


nickevansuk:
1) Improve `flow.C1()`:
- To accept request template file as a parameter, to allow some test suites to customise their request templates
- Note that some tests will book more than one event, so we will need to account for multiple OrderItems (see comment https://github.com/openactive/openactive-test-suite/issues/34#issuecomment-589641493)
- Include validation on the output of C1 with the correct validation mode (and if status code is not `200` then use "OpenBookingError" validation mode)

2) Create something like `state.getRandomEvent(["ScheduledSession", "SecheduledSession", "Slot"])` - note this is named generically with the specified type of event passed through to the microservice, preparing the way for https://github.com/openactive/openactive-test-suite/issues/21 - so we can create a suite of tests that works for random stuff specifically. Note passing an array of event types also allows for the creation of more than one event. In fact - perhaps even better - should such an array actually live in [this file](https://github.com/openactive/openactive-test-suite/blob/enhanced-logging/packages/openactive-integration-tests/test/flows/book-and-cancel/book-and-customer-cancel-success/book-free-test.js#L8) with a `randomEvent: ["ScheduledSession", "SecheduledSession", "Slot"]` array used in place of the `event` array? That way `state.createEvent(testEvent)` (see 4 below) could do both random and explicit modes depending on `testEvent`?

3) Ensure that flow.C1 pattern above extended to all endpoints (https://www.openactive.io/open-booking-api/EditorsDraft/#paths-and-verbs), noting the special treatment of `OrdersFeed` that we already have in the code

4) Ensure that the file https://github.com/openactive/openactive-test-suite/blob/enhanced-logging/packages/openactive-integration-tests/test/flows/book-and-cancel/book-and-customer-cancel-success/book-nonfree-test.js can include different `@type` (e.g. "SchuduledSession") - which it should do without issues. So perhaps we simply rename `createScheduledSession()` to be more generic (e.g. `createEvent()`), as it should handle whichever type is specified in the [file](https://github.com/openactive/openactive-test-suite/blob/enhanced-logging/packages/openactive-integration-tests/test/flows/book-and-cancel/book-and-customer-cancel-success/book-free-test.js#L10).

5) Generalise _some_ of C1, C2 tests etc to provide shared examples that can be reused across shared test suites.

nickevansuk:
@ylt note that I've expanded the above with some additional ideas, and also to include booking more than one event. To make this work within the existing framework, how about:

[...]

PR #37, opened 2020-02-25
Use JS rather than Mustache templated JSON for templates

ylt:
It doesn't make much sense to template JSON when it's already a data structure that can be programatically generated.

By generating the structures as code it allows for more dynamic structures, such as loops, etc. i.e. as required in #34


ylt:
This PR is based on a rather early version of the code, so fair few merge conflicts to resolve.

PR #45, opened 2020-03-05
Make response checking in broker more robust

ylt:
Sometimes I get random DNS failures, these would crash the broker as in these cases there's no response object.

Example printed error after this fix:
```
Error for RPDE page "https://yyyyy-test.azurewebsites.net/feeds/scheduled-sessions?afterTimestamp=2567942183&afterId=00000000001040301313": Error: getaddrinfo ENOTFOUND yyyyy-test.azurewebsites.net. Response: undefined
```

### DIFF
--- a/packages/openactive-broker-microservice/app.js
+++ b/packages/openactive-broker-microservice/app.js
@@ -95,6 +95,10 @@ function getRPDE(url, cb) {
 
 
       cb(json);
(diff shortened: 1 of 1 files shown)


nickevansuk:
Note have actually included this fix in c6cb20b, great idea!

Issue #91, opened 2020-07-15
Introduce Gradual Typing with TypeScript


lukehesluke:
Static typing is a great way to improve the reliability of code (once it's passed the type checker!). TypeScript is a particularly powerfull ([and popular](https://visualstudiomagazine.com/articles/2020/03/04/redmonk-feb20.aspx)) tool for achieving this.

It has two especially useful properties that make it a great candidate for adding to projects which have already been written in Native (untyped) JS:

- **Gradual typing**. The entire project doesn't need to be converted in one go. It can be done one file at a time or even one function at a time
- **Can use JSDoc annotations** . Rather than using the TypeScript language (which is somewhat less supported by tooling e.g. the Node.js REPL; and requires a build step), Typescript can instead use JSDoc annotations on native JS. This means that the code base does not need to change radically. In this way, type annotations also serve as improved documentation in a way that JS tools (e.g. editors) can already understand. VSCode is an example of an editor that can happily use the annotations to provide on-line type checking.

I propose typing one existing test as a way to get us started. Then, we can type any future tests using the same pattern and types.

PR #99, opened 2020-07-16
chore: Introduce TypeScript

lukehesluke:
TS is currently only set-up to check one test, but it's a start

1st part of https://github.com/openactive/openactive-test-suite/issues/91

I have not yet typed the core functions. This PR includes some work to fix or get around issues which hinder static analysis by TS (even with no types set). Mainly this is globally defined variables introduced in a few bits of the code


packages/openactive-integration-tests/ARCHITECTURE.md
```diff
@@ -83,7 +83,7 @@ Then use it within the describe blocks as so:
 
 ```jsx
 (new GetMatch({
-  state, flow, logger, configuration,
+  state, flow, logger,
```
lukehesluke, line 86: This field is unused by the `GetMatch` constructor

packages/openactive-integration-tests/test/features/core/availability-check/implemented/availability-confirmed-test.js
```diff
@@ -1,16 +1,11 @@
 /* eslint-disable no-unused-vars */
 const chakram = require('chakram');
 const chai = require('chai'); // The latest version for new features than chakram includes
-const { RequestState } = require('../../../../helpers/request-state');
```
lukehesluke, line 4: These dependencies are unused in the test. Lint didn't capture this 🤔 . Problem for another PR perhaps

packages/openactive-integration-tests/test/helpers/request-state.js
```diff
@@ -78,11 +79,13 @@ class RequestState {
           return x =>
           (Array.isArray(x.availableChannel) && x.availableChannel.includes("https://openactive.io/OpenBookingPrepayment"))
           && x.advanceBooking != "https://openactive.io/Unavailable"
+          // @ts-expect-error startDate is not defined
           && (x.validFromBeforeStartDate && moment(startDate).subtract(moment.duration(x.validFromBeforeStartDate)).isAfter());
```
lukehesluke, line 83: I'm very confused by this. My current theory is that this code branch happens to never gets executed as it would surely fail if it did. I did some investigation but stopped as it was taking a while.

Same goes for the below 2 `@ ts-expect-error`s

packages/openactive-integration-tests/test/shared-behaviours/get-match.js
```diff
@@ -2,7 +2,7 @@ const {expect} = require("chakram");
 const sharedValidationTests = require("./validation");
 
 class GetMatch {
-  constructor ({state, flow, logger, configuration, orderItemCriteria}) {
+  constructor ({state, flow, logger, orderItemCriteria}) {
```
lukehesluke, line 5: This param is unused

packages/openactive-integration-tests/ARCHITECTURE.md
nickevansuk: Yep, can be removed 👍

packages/openactive-integration-tests/test/features/core/availability-check/implemented/availability-confirmed-test.js
nickevansuk: /* eslint-disable no-unused-vars */ disables the check for them being unused. The idea here was just to have a bunch of boilerplate dependencies included in each test... though these could actually be consolidated into a helper / index which is probably the better approach here?

lukehesluke: Hah! I think I stumbled over the triple negative "disable no unused vars"

packages/openactive-integration-tests/test/helpers/request-state.js
nickevansuk: This is indeed a bug, and I've fixed this but not committed the code - doh! Will do in a separate PR

packages/openactive-integration-tests/test/shared-behaviours/get-match.js
nickevansuk: Looks good 👍

packages/openactive-integration-tests/test/features/core/availability-check/implemented/availability-confirmed-test.js
lukehesluke: @nickevansuk I personally agree with the ESLint rule for this. If we allow extraneous dependencies and no way to check, the code could end up getting messy.

Consolidation of dependencies into an index seems like an unnecessary layer of indirection to me. (One thing that I think is good to optimise is how many files a developer has to click through in order to look into something).

With VSCode (aided by TS), it should be possible to just type `RequestState` and the import automatically be added to the top of the file

packages/openactive-integration-tests/test/helpers/request-state.js
nickevansuk: @lukehesluke here's the PR for this: https://github.com/openactive/openactive-test-suite/pull/100

packages/openactive-integration-tests/test/features/core/availability-check/implemented/availability-confirmed-test.js
nickevansuk: Sure sounds good to me!

packages/openactive-integration-tests/test/helpers/request-state.js
lukehesluke: Excellent

packages/openactive-integration-tests/package.json
```diff
@@ -5,7 +5,7 @@
   "main": "app.js",
   "scripts": {
     "start": "echo \"Error: package contains only tests\" && exit 1",
-    "test": "jest",
+    "test": "tsc && jest",
```
nickevansuk, line 8: Given that "test" is the command test suite users will use, do we want to have a separate script for this? We can then include it in the GitHub Actions CI script as an explicit step?

packages/openactive-integration-tests/test/helpers/logger.js
```diff
@@ -223,24 +232,6 @@ class BaseLogger {
     else return "passed";
   }
 
-  get numValidationStatuses () {
```
nickevansuk, line 226: Is this being removed as it's unused?

lukehesluke: It is unused and it is a get accessor that returns no value, which TypeScript, by default, does not allow

packages/openactive-integration-tests/package.json
lukehesluke: Ah, good point @nickevansuk. I'll do that

lukehesluke: This is somewhat counter to the usual node.js pattern for packages (where `npm test` is expected by CI tools to check that the repo can be published e.g. run lint, run unit tests, etc) but I'm happy to keep with the established pattern

lukehesluke: @nickevansuk I've moved it to a separate `static-checks` script. I'll set it to run in a GitHub action in a separate PR

nickevansuk: Interesting... do we want to put the tests under `npm start` and put static checks under `npm test`, would that be clearer?

packages/openactive-integration-tests/test/helpers/logger.js
nickevansuk: Could we just add `return result;` for now until we get more into this? I'm suspecting a bug here where it should be used but isn't...

packages/openactive-integration-tests/package.json
lukehesluke: This would mean that the checks would be performed by any CI automatically without need of CI-specific config like GitHub actions. I think I agree with this. I can do it in another PR?

packages/openactive-integration-tests/test/helpers/logger.js
lukehesluke: Sure. This is done

packages/openactive-integration-tests/package.json
nickevansuk: Yep, perfect - and we'll need to update the docs too as it's a breaking change

PR #103, opened 2020-07-16
feat: Microservice status endpoint

nickevansuk:
Add a status endpoint to inspect the harvesting progress of the microservice


packages/openactive-broker-microservice/app.js
```diff
@@ -91,28 +98,36 @@ function getRPDE(url, headers, cb) {
         throw `Base URL of RPDE 'next' property ("${getBaseUrl(json.next)}") does not match base URL of RPDE page ("${url}")`;
       }
 
+      context.currentPage = url;
```
lukehesluke, line 101: I think a lot of mistakes and issues can be avoided by never mutating a function's arguments

For example, if you added some logic under the `setTimeout` statement that calls `getRpde()`, and that logic includes an `await`, after the `await`, context may or may not have been mutated by the subsequent calls to `getRpde()`. This may be fine now, but I would say it's quite reasonable for a programmer to not expect that sort of thing to happen, leading to potentially difficult bugs down the road and - if not bugs - more complexity that a programmer has to keep in their mind as to what can happen and what to be careful of.

Therefore, I think we should stick to a rule of never mutating a function's args (I believe the airbnb lint also has a rule for this).

In this case, I would recommend making a new `context` variable that copies the `context` in the function's arguments and adds whatever else needs to be added

packages/openactive-broker-microservice/app.js
```diff
@@ -91,28 +98,36 @@ function getRPDE(url, headers, cb) {
...
+          if (VERBOSE) console.log(`Sleep mode poll for RPDE feed "${url}"`);
         }
-        setTimeout(x => getRPDE(url, headers, cb), 500);
+        context.sleepMode = true;
+        if (context.timeToHarvestCompletion === undefined) context.timeToHarvestCompletion = millisToMinutesAndSeconds(new Date() - startTime);
+        setTimeout(x => getRPDE(url, context, cb), 500);
       } else {
-        cb(json, headers);
+        context.responseTimes.push(response.elapsedTime);
+        if (context.responseTimes.length > 5) context.responseTimes.shift();
```
lukehesluke, line 113: Comment to explain that you're maintaining a "last 5 items" buffer?

packages/openactive-broker-microservice/app.js
```diff
@@ -213,6 +228,30 @@ app.get("/dataset-site", function(req, res) {
   res.send(datasetSiteJson);
 });
 
+function mapToObject(map) {
+  if (map instanceof Map) {
+    return Object.assign(Object.create(null), ...[...map].map(v => (typeof v[1] === 'object' && v[1].size === 0 ? {} : { [v[0]]: mapToObject(v[1]) })));
+  } else if (map instanceof Set) {
+    return map.size;
```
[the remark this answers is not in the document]
nickevansuk: Good point, will add comment. This is to output the size of the set in the status endpoint - rather than the set contents itself.

packages/openactive-broker-microservice/app.js
nickevansuk: Great point, will add

packages/openactive-broker-microservice/app.js
```diff
@@ -14,10 +14,12 @@ var DATASET_SITE_URL = config.get("datasetSiteUrl");
 var REQUEST_LOGGING_ENABLED = config.get("requestLogging");
 var WAIT_FOR_HARVEST = config.get("waitForHarvestCompletion");
 var ORDERS_FEED_REQUEST_HEADERS = config.get("ordersFeedRequestHeaders");
+var VERBOSE = config.get("verbose");
```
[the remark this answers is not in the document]
nickevansuk: Ye was just consistency - this whole file was a quick-and-dirty approach that has slowly grown over time... will apply eslint --fix to make the whole thing more readable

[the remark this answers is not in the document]
nickevansuk: Ok have added a separate PR for this as it's gonna get messy: https://github.com/openactive/openactive-test-suite/pull/105

packages/openactive-broker-microservice/app.js
nickevansuk: Yes this was a quick-and-dirty solution. Had thought about maintaining a Map of contexts and having that get updated by this function... will amend so you can see what you think

nickevansuk: see new `feedContextMap` approach

lukehesluke: Magic 🧙

PR #105, opened 2020-07-17
chore: Add linting to microservice

nickevansuk:
Lint and fix microservice


packages/openactive-broker-microservice/app.js
```diff
@@ -1,29 +1,29 @@
-const express = require("express");
-const logger = require("morgan");
-const request = require("request");
-const axios = require("axios");
-const nSQL = require("@nano-sql/core").nSQL;
-const moment = require("moment");
-const config = require("config");
-const criteria = require("@openactive/test-interface-criteria").criteria;
+/* eslint-disable no-console */
```
lukehesluke, line 1: I'm surprised that this is needed if the lint config has `env.node: true` 🤔

packages/openactive-broker-microservice/app.js
```diff
@@ -1,29 +1,29 @@
...
+/* eslint-disable no-use-before-define */
```
lukehesluke, line 2: I think this is a good rule, but fair, one for another day!

nickevansuk: Yes agreed, it would have messed with the rest of the diff so good to introduce in a single PR

packages/openactive-broker-microservice/app.js
nickevansuk: Ye is strange...

PR #106, opened 2020-07-17
chore: Run static checks on CI

lukehesluke:
openactive-test-suite's `npm test` has been repurposed, so that this runs the static checks.

The Booking API tests are now run with `npm start`. This fits with the NPM pattern where `npm test` tests the code itself and `npm start` runs the code. In this case, the code is tests, so the language can be confusing, admittedly

### DIFF
--- a/.github/workflows/reference-implementation.yml
+++ b/.github/workflows/reference-implementation.yml
@@ -51,18 +51,20 @@ jobs:
       run: | 
         npm install --prefix tests/packages/openactive-broker-microservice
         npm install --prefix tests/packages/openactive-integration-tests
+    - name: Run Checks on the Code (Test the Tests!)
+      run: npm test --prefix tests
     - name: Start Broker Microservice
       run: | 
         NODE_CONFIG='{"waitForHarvestCompletion": true, "datasetSiteUrl": "https://localhost:5001/openactive"}' npm start --prefix tests/packages/openactive-broker-microservice &
     - name: Run OpenActive Integration Tests in Random Mode
       run: |
         rm -rf ./tests/packages/openactive-integration-tests/output/
-        NODE_CONFIG='{ "sellers": { "primary": { "@type": "Organization", "@id": "https://localhost:5001/api/identifiers/sellers/0", "requestHeaders": { "X-OpenActive-Test-Client-Id": "test", "X-OpenActive-Test-Seller-Id": "https://localhost:5001/api/identifiers/sellers/0" } }, "secondary": { "@type": "Person", "@id": "https://localhost:5001/api/identifiers/sellers/1" } }, "useRandomOpportunities": true, "generateConformanceCertificate": true, "conformanceCertificateId": "https://openactive.io/openactive-test-suite/example-output/random/certification/" }' npm test --prefix tests/packages/openactive-integration-tests --runInBand -- test/features/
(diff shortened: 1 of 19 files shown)


package.json
```diff
@@ -8,7 +8,9 @@
     "install": "npm install --prefix packages/openactive-broker-microservice && npm install --prefix packages/openactive-integration-tests",
     "prettier": "prettier --check \"packages/**/*.js\" \"packages/**/*.json\"",
     "prettier:write": "prettier --write \"packages/**/*.js\" \"packages/**/*.json\"",
-    "start": "npm start --prefix packages/openactive-broker-microservice",
+    "start": "echo 'Error: please select a package to run and then run it e.g. `npm start --prefix packages/openactive-broker-microservice`' && exit 1",
```
nickevansuk, line 11: `npm start` in the root could run `run-specific-test.sh`? Is that what you meant by collating?

package.json
```diff
@@ -8,7 +8,9 @@
...
+    "start-broker": "npm start --prefix packages/openactive-broker-microservice",
+    "start-tests": "npm start --prefix packages/openactive-integration-tests",
     "test": "npm test --prefix packages/openactive-integration-tests",
```
nickevansuk, line 14: Similarly `npm test` could run `npm test --prefix tests` ?

package.json
lukehesluke: @nickevansuk yeah that makes sense. I hadn't completed my PR description, sorry.

I wanted to have an `npm start` that could be run by developer or CI alike, which would just do the right thing based on the environment (e.g. config files). But that doesn't work for CI because there's scenarios like (1). Run with randomized (2). Run with controlled, which would require a way of specifying, in the environment, multiple runs - perhaps each with distinct configuration - which run with the microservice started before and stopped after all runs have completed.

Therefore, that's not possible right now without a design change. I like your suggestion though. It's at least a good scenario for a developer. Maybe just running the post-install section of `simple-ci.sh`, which just runs all the tests

```sh
# Start broker microservice in the background
npm start --prefix packages/openactive-broker-microservice &
pid=$!

# Kill broker microservice in case of error
trap 'err=$?; echo >&2 "Exiting on error $err"; kill $pid; exit $err' ERR

# Run tests
npm start --prefix packages/openactive-integration-tests

# Kill broker microservice on success
kill $pid
```

package.json
lukehesluke: @nickevansuk I'm not quite sure what you mean here. `npm test --prefix tests` would not run anything

.github/workflows/reference-implementation.yml
```diff
@@ -51,18 +51,20 @@ jobs:
       run: | 
         npm install --prefix tests/packages/openactive-broker-microservice
         npm install --prefix tests/packages/openactive-integration-tests
+    - name: Run Checks on the Code (Test the Tests!)
+      run: npm test --prefix tests
```
nickevansuk, line 55: Based on your other comment "`npm test --prefix tests` would not run anything", what does this do? I haven't tried the command myself...

package.json
nickevansuk: if we add `"$@"` at the end of the line `npm start --prefix packages/openactive-integration-tests` to add the flexibility of passing through the arguments, that would work right?

nickevansuk: And in fact we could then run it from CI also, if NODE_CONFIG prefix gets passed through.

For scenarios like (1). Run with randomized (2). Run with controlled, it's fine to stop and restart the microservice each time

.github/workflows/reference-implementation.yml
lukehesluke: `tests` is just the directory into which openactive-test-suite is git cloned in the instance that the GH action spawns

package.json
nickevansuk: Maybe we do this as above for `npm start` but leave the GitHub Actions CI as-is for now? Only because separating out the microservice from the integration tests is useful to debug where the issues are coming from - and it seems to filter the output from each rather than combining the output together (which makes the console harder to read)

.github/workflows/reference-implementation.yml
nickevansuk: Doh sorry, of course!

package.json
nickevansuk: Updated based on other comment... can we have it run both package tests from `npm test` here?

package.json
lukehesluke: tenor

package.json
lukehesluke: Yup. That's now done 👍

package.json
lukehesluke: @nickevansuk I'm going to do that on a separate branch (https://github.com/openactive/openactive-test-suite/pull/107, incomplete)

PR #107, opened 2020-07-17
chore: Start the broker and run the tests with project's `npm start`

lukehesluke:
Relies on: https://github.com/openactive/openactive-test-suite/pull/106, so this should be merged first


nickevansuk:
Looks great - only comment is the naming of `run-all-tests.sh`, as it doesn't necessarily run all the tests. Perhaps `test-with-broker.sh` or something?

nickevansuk:
Ah also CI is failing... I thought we'd agreed not to make changes to `.github/workflows/reference-implementation.yml`? Maybe I misunderstood the gif at https://github.com/openactive/openactive-test-suite/pull/106#discussion_r456585130?

lukehesluke:
Ah, the comments didn't refresh. I was looking at the comment prior to that one when I sent the gif, @nickevansuk , so I had not seen that last comment

nickevansuk:
Curiously the failing CI is now due to the `kill` apparently not working on error (the trap failing). `trap` is a `bash` thing, so maybe something in the CI env going on

nickevansuk:
Ah! Strange, GitHub's usually pretty good at updating the page...

PR #115, opened 2020-07-28
chore: Generate docs as a precommit hook

lukehesluke:
So that we don't have to remember to do this for new tests

## Demonstration

Screenshot 2020-07-28 at 17 44 25

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,7 +11,8 @@
     "start-tests": "npm start --prefix packages/openactive-integration-tests",
     "install": "cd packages/openactive-broker-microservice && npm install && cd ../openactive-integration-tests && npm install && cd ../..",
     "test": "npm test --prefix packages/openactive-broker-microservice && npm test --prefix packages/openactive-integration-tests",
-    "certificate-validator": "npm run certificate-validator --prefix packages/openactive-integration-tests"
+    "certificate-validator": "npm run certificate-validator --prefix packages/openactive-integration-tests",
+    "doc-gen": "cd packages/openactive-integration-tests && npm run doc-gen"
   },
   "dependencies": {
     "node-cleanup": "^2.1.2"
@@ -24,7 +25,7 @@
   },
   "husky": {
     "hooks": {
-      "pre-commit": "npm run doc-gen --prefix packages/openactive-integration-tests"
+      "pre-commit": "npm run doc-gen"
     }
   }
 }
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -18,5 +18,13 @@
   },
   "engines": {
     "node": "12.18.2"
+  },
+  "devDependencies": {
+    "husky": "^4.2.5"
+  },
+  "husky": {
+    "hooks": {
+      "pre-commit": "npm run doc-gen --prefix packages/openactive-integration-tests"
```
nickevansuk, line 27: So it turns out that "--prefix packages/openactive-integration-tests" doesn't run on Windows, which hinders development on Windows if this is in a hook. Suggest replacing this with a `cd` prefix similar to https://github.com/openactive/openactive-test-suite/pull/116

lukehesluke: @nickevansuk is it just `npm install` which doesn't work with `--prefix` on windows? e.g. the only [bug](https://github.com/npm/cli/issues/1290) [reports](https://html.developreference.com/article/11036084/NPM+prefix+flag+does+not+work+as+expected+on+windows) [I can find](https://github.com/npm/npm/issues/20245) relate to `npm install` and I see that we've only removed `--prefix` from our `install` script. Our test script still looks like `npm test --prefix packages/openactive-broker-microservice ...`

lukehesluke: My apologies, I have now seen and commented on your other PR (https://github.com/openactive/openactive-test-suite/pull/116), which changes all of the scripts

nickevansuk: No worries - based on my testing it is mainly `install` that has the issue, but it appears to be an anti-pattern more generally? - was looking for a reference in the npm docs or repos where `--prefix` is recommended for this purpose and couldn't find one... 

If we choose to use `cd` here, would be good to check that husky uses a sub-shell like `npm` does (see https://github.com/openactive/openactive-test-suite/pull/116#issuecomment-665567302), to ensure there's no issue with the working dir restore.

If not perhaps we add an npm script in the root that calls `npm run doc-gen` in the packages dir using `cd`, then just call this from `husky`? Perhaps better taking this approach anyway for consistency?

lukehesluke: > If not perhaps we add an npm script in the root that calls npm run doc-gen in the packages dir using cd, then just call this from husky? Perhaps better taking this approach anyway for consistency?

👍  good idea

lukehesluke: @nickevansuk this is now done

PR #116, opened 2020-07-28
fix: Remove prefix anti-pattern

nickevansuk:
Remove the `--prefix` anti-pattern as it does not work on Windows, and replace with `cd`, as discussed in https://github.com/npm/npm/pull/10958

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,11 +7,15 @@
   "scripts": {
     "start": "node start.js",
     "debug": "node --inspect start.js -- --runInBand",
-    "start-broker": "npm start --prefix packages/openactive-broker-microservice",
-    "start-tests": "npm start --prefix packages/openactive-integration-tests",
-    "install": "cd packages/openactive-broker-microservice && npm install && cd ../openactive-integration-tests && npm install && cd ../..",
-    "test": "npm test --prefix packages/openactive-broker-microservice && npm test --prefix packages/openactive-integration-tests",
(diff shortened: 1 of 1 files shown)


lukehesluke:
If a command fails, then the user will have changed directory to the sub-folder without reverting the directory change.

Why is `--prefix` an anti-pattern? The NPM issue referenced isn't discussing `--prefix` but some proposed extra context to add to an NPM script

nickevansuk:
The assumption of this being an anti-pattern was based on:
- It looks like `--prefix` is [designed to modify the dir used for install](https://docs.npmjs.com/misc/config#prefix), rather than change the working directory for other operations.
- In the [issue above](https://github.com/npm/npm/pull/10958) the maintainers seem to be recommending using `cd` to update the working directory, instead of using `--prefix`.
- The issue above also clarifies that npm scripts are executed in a sub-shell, and so the previous working directory is restored after the sub-shell exits. Hence the `cd` prefix being recommended. My testing on Windows confirms this is the case.

Haven't seen `--prefix` being used widely for the purpose of updating the working directory, though interested if we've come across anything that does use this?

lukehesluke:
> The issue above also clarifies that npm scripts are executed in a sub-shell, and so the previous working directory is restored after the sub-shell exits. Hence it recommending the cd approach. My testing on Windows confirms this is the case.

Fair! I missed that. I was worried that this would mess with CI. In which case, I agree and I'll update my other PR

PR #117, opened 2020-07-28
docs: Clarify README for deployment config

nickevansuk:
Make config documentation more helpful

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ Before running, configure the test suite:
 
 ### Local configuration
 
-A `local.json` file can be placed alongside each `default.json`, and that the subset of properties defined within it will override those in `default.json`. It is recommended that for development and deployment a `local.json` file is created instead of making changes to the `default.json` file, so that any new required settings that are added in future versions can be automatically updated in `default.json`.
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -23,6 +23,16 @@ Before running, configure the test suite:
  - `packages/openactive-integration-tests/config/default.json`
    - [More information](./packages/openactive-integration-tests/#configuration)
 
+### Local configuration
+
+A `local.json` file can be placed alongside each `default.json`, and that the subset of properties defined within it will override those in `default.json`. It is recommended that for development and deployment a `local.json` file is created instead of making changes to the `default.json` file, so that any new required settings that are added in future versions can be automatically updated in `default.json`.
```
lukehesluke, line 28:
```suggestion
A `local.json` file can be placed alongside each `default.json`, and the subset of properties defined within it will override those in `default.json`. It is recommended that for development and deployment a `local.json` file is created instead of making changes to the `default.json` file, so that any new required settings that are added in future versions can be automatically updated in `default.json`.
```

nickevansuk: Great thanks!

PR #119, opened 2020-07-29
chore: Type the doc generator + improve feature helper types

lukehesluke:
I was trying to understand how the "criteria required" parts of the docs were generated. Thought I might as well type the bits as I understand them

### DIFF
--- a/packages/openactive-integration-tests/tsconfig.json
+++ b/packages/openactive-integration-tests/tsconfig.json
@@ -5,9 +5,11 @@
     "checkJs": true,
     "downlevelIteration": true,
     "target": "ES2015",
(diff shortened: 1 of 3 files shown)


packages/openactive-integration-tests/tsconfig.json
```diff
@@ -5,9 +5,11 @@
     "checkJs": true,
     "downlevelIteration": true,
     "target": "ES2015",
-    "moduleResolution": "node"
+    "moduleResolution": "node",
+    "resolveJsonModule": true
```
lukehesluke, line 9: This is required for the `require('../package.json')` call in the doc generator

PR #122, opened 2020-07-31
feat(amending-order-quote): Create amend-c1 test

lukehesluke:
Additionally, add a note to the openactive-integration-tests README for how to debug individual tests within VSCode


packages/openactive-integration-tests/documentation/generator.js
```diff
@@ -229,13 +229,14 @@ function renderFeatureTest(t) {
...
  */
-function renderCriteriaRequired(c, prefix) {
-  if (c.size == 0) {
+function renderCriteriaRequired(criteriaRequired, prefixOverride) {
+  if (criteriaRequired.size == 0) {
     return '';
   } else {
-    return `${prefix !== undefined ? prefix : '\nPrerequisite opportunities per Opportunity Type: '}${Array.from(c.entries()).map(([key, value]) => `[${key}](https://openactive.io/test-interface#${key}) x${value}`).join(', ')}`;
+    const prefix = prefixOverride !== undefined ? prefixOverride : '\nPrerequisite opportunities per Opportunity Type: ';
+    return `${prefix}${Array.from(criteriaRequired.entries()).map(([key, value]) => `[${key}](https://openactive.io/test-interface#${key}) x${value}`).join(', ')}`;
```
lukehesluke, line 240: I thought I was going to update this function though I didn't need to in the end. I made it a tiny bit tidier anyway 🤷

packages/openactive-integration-tests/package.json
```diff
@@ -35,15 +35,19 @@
...
     "jest-spec-reporter": "^1.0.10",
     "jszip": "^3.5.0",
+    "lodash": "^4.17.19",
     "mkdirp": "^1.0.3",
     "moment": "^2.24.0",
     "p-memoize": "^3.1.0",
     "rmfr": "^2.0.0",
-    "strip-ansi": "^6.0.0"
+    "strip-ansi": "^6.0.0",
+    "uuid": "^8.3.0"
```
lukehesluke, line 44: before, we seemed to have been using an implicitly installed version of uuid, which was in a separate package. This is bad as if that package privately updates it's dependencies so that it no longer uses uuid, this project would crash.

Intriguing that lint didn't catch that 🤔

packages/openactive-integration-tests/test/features/README.md
```diff
@@ -20,6 +20,7 @@ The tests for these features provide partial coverage but do not include all kno
 
 | Category | Feature | Specification | Description | Prerequisites per Opportunity Type |
 |----------|---------|---------------|-------------|-------------------|
+| core | Amending the OrderQuote before B ([amending-order-quote](./core/amending-order-quote/README.md)) | [Required](https://www.openactive.io/open-booking-api/EditorsDraft/#amending-the-orderquote-before-b) | Allows the basket to be updated for a particular order | [TestOpportunityBookable](https://openactive.io/test-interface#TestOpportunityBookable) x8 |
```
lukehesluke, line 23: Note that it's `x8` due to the `numOpportunitiesUsedPerCriteria: 2`

packages/openactive-integration-tests/test/helpers/feature-helper.js
```diff
@@ -8,33 +9,32 @@ const BOOKABLE_OPPORTUNITY_TYPES_IN_SCOPE = global.BOOKABLE_OPPORTUNITY_TYPES_IN
...
+ * @property {boolean} [runOnce]
+ * @property {boolean} [skipMultiple]
+ * @property {boolean} [runOnlyIf]
+ * @property {number} [numOpportunitiesUsedPerCriteria] How many opportunities
+ *   are used by the test per criteria. e.g. if each test iteration needs to
+ *   fetch 2 opportunities, this number should be 2.
+ *
+ *   Used to generate the docs for each test.
+ *
+ *   Defaults to 1.
```
lukehesluke, line 37: This is another (more verbose) way to specify an object type in TypeScript-in-JSDoc. It allows having docs for individual fields, as I have introduced for `numOpportunitiesUsedPerCriteria`

packages/openactive-integration-tests/test/helpers/flow-helper.js
```diff
@@ -1,40 +1,51 @@
...
+ * Additionally, flow functions call any dependent functions (e.g. B calls C2,
+ * which calls C1, etc).
+ */
 class FlowHelper {
   constructor (helper) {
     this.state = helper;
   }
 
-  getDatasetSite = pMemoize(async getDatasetSite => {
+  getDatasetSite = pMemoize(async () => {
```
lukehesluke, line 18: I checked all the uses of these functions, and none of them supply an arg to any of these flow functions. And these args are not used within the function itself. Therefore, they are useless. I suspect they were introduced as a misunderstanding of async arrow function syntax

packages/openactive-integration-tests/test/helpers/request-state.js
```diff
@@ -23,15 +27,16 @@ function isResponse(response) {
 }
 
 class RequestState {
-  constructor (logger) {
+  /**
+   * 
+   * @param {InstanceType<import('./logger')['Logger']>} logger 
+   * @param {string | null} uuid
```
lukehesluke, line 33: It is now possible to instantiate a RequestState with a `uuid`. See my comment in `amend-c1-test.js` to see why

packages/openactive-integration-tests/test/types/OpportunityCriteria.d.ts
```diff
@@ -0,0 +1,9 @@
+export interface OpportunityCriteria {
+  opportunityType: string | null;
+  opportunityCriteria: string;
+  primary: boolean;
+  control: boolean;
+  opportunityReuseKey?: number;
+  usedInOrderItems?: number;
+  seller?: string;
+}
```
lukehesluke, line 9: Here I'm using a `.d.ts` file to specify a type within TypeScript itself. This type is now used across multiple files so it didn't make sense to localize it only in feature-helper

packages/openactive-integration-tests/test/features/core/amending-order-quote/implemented/amend-c1-test.js
```diff
@@ -0,0 +1,106 @@
...
+          'orderedItem.@id': feedOrderItem.orderedItem['@id'],
+        });
+      });
+  };
+  // N.B.: The following two tests must be performed sequentially - with
+  // Second Attempt occurring after First Attempt.
+  describe('First Attempt - C1', () => {
+    // Each of these scenarios uses a separate state and flowHelper because they fetch separate opportunities
+    const state = new RequestState(logger, uuid);
+    const flow = new FlowHelper(state);
```
lukehesluke, line 47: A new state and flow are used in each "attempt" because both attempts use different sets of opportunities. They are linked by the UUID to ensure they use the same Order

I thought this was much simpler and clearer than introducing some functionality to state and flow to support multiple batches or to empty their contents but I await your thoughts!

packages/openactive-integration-tests/documentation/generator.js
nickevansuk: 👍

packages/openactive-integration-tests/package.json
nickevansuk: great catch!

packages/openactive-integration-tests/test/features/README.md
nickevansuk: 👍

packages/openactive-integration-tests/test/features/core/amending-order-quote/implemented/amend-c1-test.js
```diff
@@ -0,0 +1,106 @@
...
+const { FlowHelper } = require('../../../../helpers/flow-helper');
+const { RequestState } = require('../../../../helpers/request-state');
+const { generateUuid } = require('../../../../helpers/generate-uuid');
+
+FeatureHelper.describeFeature(module, {
+  testCategory: 'core',
+  testFeature: 'amending-order-quote',
+  testFeatureImplemented: true,
+  testIdentifier: 'amend-c1',
+  testName: 'Amend an already-made C1 request',
```
nickevansuk, line 14:
```suggestion
  testName: 'Amend an existing OrderQuote',
```

As that's what's actually happening?

packages/openactive-integration-tests/test/features/core/amending-order-quote/implemented/amend-c1-test.js
nickevansuk: Yep this looks good - agreed, cleaner this way

packages/openactive-integration-tests/test/features/core/amending-order-quote/implemented/amend-c1-test.js
```diff
@@ -0,0 +1,106 @@
...
+      state, flow, logger,
+    }))
+      .beforeSetup()
+      .successChecks()
+      .validationTests();
+
+    // This should verify that the order has now been amended
+    itFeedItemAndResponseItemShouldMatchIds(state, c1, () => state.c1Response);
+
+    const b = (new B({
```
nickevansuk, line 96: C2 is required before calling B...

Perhaps this test should actually go C1 -> C2 -> C1 -> C2 -> B?

packages/openactive-integration-tests/test/features/core/amending-order-quote/implemented/amend-c1-test.js
```diff
@@ -0,0 +1,106 @@
...
+      state, flow, logger,
+    }))
+      .beforeSetup()
+      .successChecks()
+      .validationTests();
+
+    itFeedItemAndResponseItemShouldMatchIds(state, c1, () => state.c1Response);
+  });
+  /** Fetch some new opportunities and amend the existing order at C1, and then complete it */
+  describe('Second Attempt - C1 -> B', () => {
```
nickevansuk, line 70: Notice this block duplicates the code of the previous attempt - could we pull this out into a function?

packages/openactive-integration-tests/test/helpers/flow-helper.js
nickevansuk: There used to be a thing about naming functions so they could be easily recognised in a stack trace...? Not sure if that guidance is now out-of-date? If that's not it this looks great!

lukehesluke: > There used to be a thing about naming functions so they could be easily recognised in a stack trace...?

The above doesn't name the function. It acts as the first agument to the function. Node.js stack traces for promises and arrow functions are much better than they used to be, so I believe `getSatasetSite` would be picked up by stack traces though I'm not completely sure.

packages/openactive-integration-tests/test/features/core/amending-order-quote/implemented/amend-c1-test.js
lukehesluke: > C2 is required before calling B...

Yes. The flow helper calls C2 if it hasn't already been called though. Are you saying that it's better to be explicit and call it in the test?

> Perhaps this test should actually go C1 -> C2 -> C1 -> C2 -> B?

Are you saying that the 1st attempt should call C2 after C1? In the spreadsheet, that's a separate test

packages/openactive-integration-tests/test/features/core/amending-order-quote/implemented/amend-c1-test.js
lukehesluke: We could do. I normally use the Rule of Three for small things

packages/openactive-integration-tests/test/features/core/amending-order-quote/implemented/amend-c1-test.js
lukehesluke: Yes, good point! I renamed it to "Amend, at C1, an existing OrderQuote" to differentiate it from the other tests for this feature (e.g. another test amends at C2)

packages/openactive-integration-tests/test/helpers/flow-helper.js
nickevansuk: sorry you're totally right! in that case lgtm!

packages/openactive-integration-tests/test/features/core/amending-order-quote/implemented/amend-c1-test.js
nickevansuk: Perfect

packages/openactive-integration-tests/test/features/core/amending-order-quote/implemented/amend-c1-test.js
nickevansuk: > flow helper calls C2 if it hasn't already been called
I wasn't aware of this, in which case no worries!

And sorry I should have checked the sheet, yes a separate test would indeed be better, and is needed if it's already in the sheet.

packages/openactive-integration-tests/test/features/core/amending-order-quote/implemented/amend-c1-test.js
nickevansuk: Sure, perhaps within the tests is good to be a little more concise for maintainability and clarity...?

packages/openactive-integration-tests/test/features/core/amending-order-quote/implemented/amend-c1-test.js
lukehesluke: I've also made the C2 call explicit. I agree that we should move towards the tests being understandable even if you don't understand the framework. This also means that changes to the framework have less destabilising effects on the tests

PR #124, opened 2020-08-03
fix(simple-book-free-opportunities): Rename with-payment-property-test -> without-payment-property-test

lukehesluke:
It was incorrectly named, meaning e.g. that the link in the simple-book-free-opportunities/README didn't work as the test identifier doesn't match the test name


nickevansuk:
Great spot!

PR #126, opened 2020-08-04
fix(amending-order-quote/amend-c1): Delete Order after 2nd attempt

lukehesluke:
I saw that deleting the order was mentioned in `DEVELOPMENT.md` and in the tests that made actual bookings e.g. `packages/openactive-integration-tests/test/features/core/simple-book-free-opportunities/implemented/without-payment-property-test.js`, so I've added it to the test that I made that makes actual bookings.

I'm curious though, is it possible that Orders will stay up on a booking system if the openactive-test-suite, say, crashes half-way through a test and is therefore unable to execute the code in `afterAll()`?

### DIFF
--- a/packages/openactive-integration-tests/test/features/core/amending-order-quote/implemented/amend-c1-test.js
+++ b/packages/openactive-integration-tests/test/features/core/amending-order-quote/implemented/amend-c1-test.js
@@ -46,8 +46,11 @@ FeatureHelper.describeFeature(module, {
    * and then fetch some opportunities and run C1
    *
    * Note: This generates jest blocks like `beforeAll()`, `it()`, etc. Therefore, this must be run within a `describe()` block
+   *
+   * @param {object} args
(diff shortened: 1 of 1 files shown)


nickevansuk:
Yes that's right. The reason that the ["DELETE" test interface endpoint](https://openactive.io/test-interface/#delete-test-interfacedatasetstestdatasetidentifier) exists is to clean up anything that was left behind from a previous failed test run.

Note also that Order Deletion is its own feature (["order-deletion"](https://github.com/openactive/openactive-test-suite/blob/master/packages/openactive-integration-tests/test/features/core/order-deletion/README.md)).

Looking at this again now, might it actually be better not to delete the Orders within each test? As, for example, it might be that an implementer wants to inspect the state of the Order following a successful test run?

Perhaps "deleteOrderOnTestSuccess" should be a config option?

lukehesluke:
What you're saying makes a lot of sense @nickevansuk . I'll close this PR for now and remove any `deleteOrder`s I come across in tests that I'm updating whenever deleting the order isn't an explicit part of the test description

PR #127, opened 2020-08-04
feat(simple-book-free-opportunities): Create with-erroneous-payment-property-test

lukehesluke:
Additionally:

- Update breq so that it can use different templates if needed (e.g. to ensure that a B request includes a `payment` property for free opportunities)
- Update without-payment-property-test so that it is always explicitly withholding the `payment` property
- Partially type chakram, which does not, unfortunately have any community TS defines


packages/openactive-integration-tests/DEVELOPMENT.md
```diff
@@ -76,16 +76,11 @@ function (configuration, orderItemCriteria, featureIsImplemented, logger, state,
 
     return chakram.wait();
   });
-
-  afterAll(async function () {
-    await state.deleteOrder();
```
lukehesluke, line 81: Based on our earlier conversation, I've removed this from the default advice for new tests

packages/openactive-integration-tests/test/features/core/simple-book-free-opportunities/implemented/with-erroneous-payment-property-test.js
```diff
@@ -0,0 +1,74 @@
...
+  testFeature: 'simple-book-free-opportunities',
+  testFeatureImplemented: true,
+  testIdentifier: 'with-erroneous-payment-property',
+  testName: 'Fail free bookings which include erroneous payment property',
+  testDescription: 'C1, C2 and B with payment property: payment property is provided but not expected in the request, so an UnnecessaryPaymentDetailsError must be returned.',
+  // The primary opportunity criteria to use for the primary OrderItem under test
+  testOpportunityCriteria: 'TestOpportunityBookableFree',
+  // The secondary opportunity criteria to use for multiple OrderItem tests
+  // This must also be TestOpportunityBookableFree as the payment property is only disallowed if ALL items are free.
+  controlOpportunityCriteria: 'TestOpportunityBookableFree',
```
lukehesluke, line 19: I just want to confirm that this makes sense. My theory so far is that the "control" is so that the "multiple order items" test has some normal opportunities as well as the specialised opportunities relevant to the test.

TODO

nickevansuk: Yep that's right, and makes sense

packages/openactive-integration-tests/DEVELOPMENT.md
nickevansuk: @lukehesluke think I mentioned before, but I wonder if we should in fact make this configurable? As it might be that systems want to minimise the number of new orders created, especially in Random mode?

lukehesluke: That makes sense. So I'll leave this for now (as we'd probably want this configuration option to be exercised in the feature-helper rather than in the tests' afterAll blocks?) and then create an issue for setting this up. Sound good?

nickevansuk: Sounds great!

lukehesluke: Issue: https://github.com/openactive/openactive-test-suite/issues/142

PR #135, opened 2020-08-13
feat: Remove non-ascii characters causing issues with console output

nickevansuk:
Issues with console output on Windows, so have reverted to only ASCII output to make this as cross-platform as possible


nickevansuk:
It really is - looks like a bug that's relatively new too, as only appeared since I updated Git for Windows. Will merge for now and we can always revert the ticks later if we figure out what the encoding fix is.

PR #139, opened 2020-08-13
feat: Refactor test-interface-criteria to allow relevant Offers to be filtered

nickevansuk:
Refactor to reduce boilerplate while centralising Offer filter logic, to remove bugs as well as increase maintainability.

Thoughts on this new approach welcome! Is it as concise as it can be?


packages/openactive-integration-tests/test/helpers/request-state.js
```diff
@@ -99,29 +99,7 @@ class RequestState {
...
-          (Array.isArray(x.availableChannel) && x.availableChannel.includes("https://openactive.io/OpenBookingPrepayment"))
-          && x.advanceBooking != "https://openactive.io/Unavailable"
-          && (x.validFromBeforeStartDate && moment(opportunity.startDate).subtract(moment.duration(x.validFromBeforeStartDate)).isAfter());
-        default: 
-          return x =>
-          (Array.isArray(x.availableChannel) && x.availableChannel.includes("https://openactive.io/OpenBookingPrepayment"))
-          && x.advanceBooking != "https://openactive.io/Unavailable"
-          && (!x.validFromBeforeStartDate || moment(opportunity.startDate).subtract(moment.duration(x.validFromBeforeStartDate)).isBefore());
-      }
-    };
```
lukehesluke, line 119: New approach is definitely a great improvement on the potential mega functions!

packages/test-interface-criteria/src/criteria/Criteria.js
```diff
@@ -13,29 +13,17 @@ module.exports = class Criteria {
...
-		
-		return { matchesCriteria, unmetCriteriaDetails };
-	}
+	get opportunityConstraints() {
+    return {};
+  }
+
+  get offerConstraints() {
+    return {}
+  }
```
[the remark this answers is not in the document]
lukehesluke: Btw, I'm almost certain that you'd have to apply this function type signature to all the "inherited" methods in the classes that extend Criteria

packages/test-interface-criteria/src/criteria/TestOpportunityBookable.js
```diff
@@ -1,35 +1,27 @@
...
 
-    if (
-      !(remainingCapacity > 0)
-    ) {
-      matchesCriteria = false;
-      unmetCriteriaDetails.push("No remaining capacity")
+  get offerConstraints() {
+    return {
+      ...super.offerConstraints,
+      'Must have "bookable" offer': (x, opportunity) =>
```
lukehesluke, line 20: Is `x` an `offer`?

packages/test-interface-criteria/src/criteria/TestOpportunityBookable.js
```diff
@@ -1,35 +1,27 @@
...
+        (Array.isArray(x.availableChannel) && x.availableChannel.includes("https://openactive.io/OpenBookingPrepayment"))
+        && x.advanceBooking != "https://openactive.io/Unavailable"
```
lukehesluke, line 22:
```suggestion
        && x.advanceBooking !== "https://openactive.io/Unavailable"
```

though perhaps the bigger thing is that eslint should be set up for test-interface-criteria

packages/test-interface-criteria/src/index.js
```diff
@@ -4,12 +4,54 @@
...
+  return opportunity.offers || (opportunity.superEvent && opportunity.superEvent.offers) || []; // Note FacilityUse does not have bookable offers, as it does not allow inheritance
+}
+
+function filterRelevantOffers(criteria, opportunity) {
+  const offers = getOffers(opportunity);
+  return Object.entries(criteria.offerConstraints)
+    .reduce((offers, [, test]) => offers.filter(offer => test(offer, opportunity)), offers);
+}
+
+function testMatch(criteria, opportunity) {
```
lukehesluke, line 17: I would personally benefit from a lil doc string for this function as I think it's rather important. It takes me a while to understand what this function does just by reading it

packages/test-interface-criteria/src/index.js
```diff
@@ -4,12 +4,54 @@
  */
 const criteria = require("./criteria").criteria.map(x => new x());
 
+function getOffers(opportunity) {
+  return opportunity.offers || (opportunity.superEvent && opportunity.superEvent.offers) || []; // Note FacilityUse does not have bookable offers, as it does not allow inheritance
+}
+
+function filterRelevantOffers(criteria, opportunity) {
```
lukehesluke, line 11: I've been quite confused reading these functions because the param name is the same as the module-level variable defined in line 5. Again, this is something ESLint would catch. You could perhaps rename the module-level variable to `allCriteria`?

packages/test-interface-criteria/src/index.js
lukehesluke: Actually nvm. I was just confused because `criteria` is the name of the param as well as the module-level variable (as mentioned in another comment)

packages/test-interface-criteria/src/criteria/Criteria.js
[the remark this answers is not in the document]
nickevansuk: Interesting! Here's some rationale, but very much welcome your thoughts!

So they're effectively returning constant values that don't need to be calculated each time they're called. Using getters reflects this intention.

Ideally we'd use [public field declarations](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes#Public_field_declarations) for this kind of thing, however they're [not here yet](https://tc39.es/proposal-class-fields/). This looks like the right optimisation when they're available though?

If optimisation was required in the meantime I was thinking either lazy getters, or some other form of memoization, e.g. using proxy handlers, or forcing the public methods of the package to accept the criteria name as a string and handling it there?), but the MVP of just using getters was easiest to start with.

Also worth noting that the same pattern is used in `ReportGenerator`.

Another option considered was to do this manually with objects instead of classes, though that would effectively be manually implementing class inheritance logic, which seems bad?

Style question:
- Do we always prefer functions for consistency, even for simple cases? ([AirBnB have this opinion](https://github.com/airbnb/javascript#accessors--no-getters-setters))

So options:
a) Leave this as getters on the basis that anything that in the future will use public field declarations will be represented as a getter for now (but change the project's style going forward so that we don't use getters and setters for other cases?)
b) Update to functions in the interim, for total consistency?
c) Use an object-based approach rather than classes?

packages/openactive-integration-tests/test/helpers/request-state.js
```diff
@@ -94,29 +94,7 @@ class RequestState {
   }
 
 	getRandomRelevantOffer(opportunity, opportunityCriteria) {
```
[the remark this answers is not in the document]
nickevansuk: Good catch, it isn't but will add

packages/test-interface-criteria/src/index.js
```diff
@@ -4,12 +4,54 @@
...
+  const offers = getOffers(opportunity);
+  return Object.entries(criteria.offerConstraints)
+    .reduce((offers, [, test]) => offers.filter(offer => test(offer, opportunity)), offers);
+}
```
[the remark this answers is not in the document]
nickevansuk: Ok great, will add!

packages/test-interface-criteria/src/criteria/Criteria.js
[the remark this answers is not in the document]
nickevansuk: Ok super, will do. Will update to Tuples, agree is more relevant, and similar syntactic sugar is available there to extend them. Presumably the tuple itself can be typed too?

packages/test-interface-criteria/src/criteria/TestOpportunityBookable.js
nickevansuk: Yes, and should be `offer`, will refactor

packages/test-interface-criteria/src/criteria/TestOpportunityBookable.js
nickevansuk: Good catch, and yes will do!

packages/test-interface-criteria/src/index.js
nickevansuk: Ah sure, though can easily add a doc string!

packages/test-interface-criteria/src/index.js
nickevansuk: Good idea, yes will sort with ESLint stuff

packages/test-interface-criteria/src/criteria/Criteria.js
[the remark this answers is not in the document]
lukehesluke: @nickevansuk thanks for explaining. I realise that I misinterpreted when I said that the result was different at different times because that logic is bound in a function. The constraint just returns the function, which is a constraint.

I would highly agree with the object-based approach. I'm a believer in only providing as much structure as is needed. With any more, it's easy to complicate things in the future. e.g. right now the point of inheritance is to extend the constraints (` { ...super.offerConstraints }`). However, inheritance also leads the way to inheriting `name`, for example, if an extended class forgets to implement `get name() { }` (which cannot be enforced in JS). Similarly, methods like `getId()` _could_ be overwritten in the future in one of the classes, meaning that it's sub-classes would now have this new functionality. Neither of these outcomes make sense - we want each criteria to have a distinct name and `getId()` is just a static function - it does not depend on which criteria.

Additionally, classes can have multiople instances, which is why you're talking about things like lazy getters or memoization (or singletons, etc). I feel like this adds a lot of entities to what are - to me - just data

Therefore, a simpler structure might be:

`Criteria.d.ts`

```ts
export type Criteria = {
  name: string,
  opportunityConstraints: {
    [ constraintName: string ]: (opportunity: any) => boolean,
  },
  offerConstraints: {
    [ constraintName: string ]: (offer: any, opportunity: any) => boolean,
  },
};
```

`criteriaUtils.js`

These are just functions:

```js
function getId(opportunity) {
  return opportunity['@id'] || opportunity['id'];
}

function getType(opportunity) {
  return opportunity['@type'] || opportunity['type'];
}
```

and then each criteria file would be like:

```js
/**
 * @type {import('./Criteria').Criteria}
 */
const TestOpportunityBookable = {
  name: 'TestOpportunityBookable',
  opportunityConstraints: { ... CriteriaFutureScheduledOpportunity.opportunityConstraints. ... },
  ...
};
```

I think the class structure makes more sense before TypeScript because you want some way of saying "these are all the same kind of thing and they all have these fields". But a TypeScript type is, I think, a much simpler way of doing that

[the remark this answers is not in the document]
lukehesluke: > Presumably the tuple itself can be typed too?

Yes. That would look like:

```ts
/**
 * - 1st item is the name of the constraint
 * - 2nd item is a predicate which checks if an opportunity satisfies that constraint
 */
type OpportunityConstraint = [string, (opportunity: any) => boolean]
type OpportunityConstraints = OpportunityConstraint[];
```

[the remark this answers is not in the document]
nickevansuk: Agreed with the object-based approach here

PR #140, opened 2020-08-14
chore: Check test-interface-criteria code using TypeScript

lukehesluke:


packages/test-interface-criteria/src/criteria/CriteriaFutureScheduledOpportunity.js
```diff
@@ -12,7 +12,7 @@ module.exports = class CriteriaFutureScheduledOpportunity extends Criteria {
     var eventStatus = opportunity.eventStatus;
 
     if (
-      !(Date.parse(startDate) > new Date(Date.now() + (3600 * 1000 * 2)))
+      !(Date.parse(startDate) > (new Date(Date.now() + (3600 * 1000 * 2))).getTime())
```
lukehesluke, line 15: Screenshot 2020-08-14 at 11 36 52

TypeScript noted this as an error. It is technically possible to compare a JS date and an integer, but it is generally recommended against doing this sort of thing (as is e.g. comparing "2" with 2 - something else that JS allows but TS does not). This might be because TS always turns on JS strict mode 🤔

nickevansuk, line 15: Good spot!

nickevansuk: (Have updated in other branch)

nickevansuk: Actually is this correct? Wouldn't both sides need `.getTime()`?

nickevansuk: Sorry ignore me, yep all good!

nickevansuk: Good catch!

Issue #137, opened 2020-08-13
Repeated Get-Match requests for availability-check opportunity-full


nickevansuk:
When testing against the EA implementation, with the config below, it appears as though the following call is repeated for both C1 and C2, instead of only being called within Get Opportunity Feed Items.

```
### Opportunity Feed extract for OrderItem 0 Request
GET http://localhost:3000/get-cached-opportunity/https%3A%2F%2Feveryoneactivebookingfacade-test.azurewebsites.net%2Fapi%2Fidentifiers%2Fscheduled-sessions%2F0792R09M1731320%2Fevents%2F2020-08-18T16%253A30%253A00Z%2F790976424
```

### Config

```json
{
  "datasetSiteUrl": "https://everyoneactivebookingfacade-test.azurewebsites.net/openactive",
  "requestLogging": false,
  "waitForHarvestCompletion": true,
  "verbose": false,
  "ordersFeedRequestHeaders": {
    "X-OpenActive-Test-Client-Id": "test"
  }
}
```

```json
{
  "useRandomOpportunities": true,
  "bookableOpportunityTypesInScope": {
    "ScheduledSession": true
  },
  "implementedFeatures": {
    "dataset-site": true,
    "availability-check": true,
    "simple-book-free-opportunities": true,
    "simple-book-with-payment": true,
    "customer-requested-cancellation": true
  },
  "sellers": {
    "primary": {
      "@type": "Organization",
      "@id": "https://everyoneactivebookingfacade-test.azurewebsites.net/api/identifiers/seller",
      "requestHeaders": {
        "X-OpenActive-Test-Client-Id": "test",
        "X-OpenActive-Test-Seller-Id": "https://everyoneactivebookingfacade-test.azurewebsites.net/api/identifiers/seller"
      }
    }
  }
}
```

## Example output

[availability-check_opportunity-full_ScheduledSession.md.txt](https://github.com/openactive/openactive-test-suite/files/5070917/availability-check_opportunity-full_ScheduledSession.md.txt)

PR #144, opened 2020-08-17
fix(flow-helper): Cache promise rejections

lukehesluke:
So that failing API calls are not made repeatedly after the 1st failure

## QA

(with configs set as in https://github.com/openactive/openactive-test-suite/issues/137):

Screenshot 2020-08-17 at 16 34 34


packages/openactive-integration-tests/test/helpers/request-state.js
```diff
@@ -1,5 +1,4 @@
 const RequestHelper = require("./request-helper");
-const pMemoize = require("p-memoize");
```
lukehesluke, line 2: unused

packages/openactive-integration-tests/test/report-generator.js
```diff
@@ -1,6 +1,5 @@
 const chalk = require("chalk");
 const Handlebars = require("handlebars");
-const pMemoize = require("p-memoize");
```
lukehesluke, line 3: unused

packages/openactive-integration-tests/test/helpers/flow-helper.js
```diff
@@ -20,32 +20,32 @@ class FlowHelper {
 
   getDatasetSite = pMemoize(async () => {
     return this.state.getDatasetSite();
-  });
+  }, { cachePromiseRejection: true });
```
lukehesluke, line 23: Very simple solution. Based on my understanding of the Test Suite, this shouldn't cause any issues and is in fact the behaviour that we want. Let me know what you think though as I'm still not too familiar with the project and may have overlooked something

nickevansuk: Yep this makes sense - we only want it to be called once and cached once within the flow

PR #146, opened 2020-08-18
chore: TypeScript broker microservice

lukehesluke:


packages/openactive-broker-microservice/app.js
```diff
@@ -2,7 +2,7 @@
 const express = require('express');
 const logger = require('morgan');
 const request = require('request');
-const axios = require('axios');
+const { default: axios } = require('axios');
```
lukehesluke, line 5: This is unfortunately required for axios types to be registered correctly due to the way that they transpile ES Modules to CommonJS

packages/openactive-broker-microservice/app.js
```diff
@@ -15,7 +15,7 @@ const WAIT_FOR_HARVEST = config.get('waitForHarvestCompletion');
 const ORDERS_FEED_REQUEST_HEADERS = config.get('ordersFeedRequestHeaders');
 const VERBOSE = config.get('verbose');
 
-process.env.NODE_TLS_REJECT_UNAUTHORIZED = 0;
+process.env.NODE_TLS_REJECT_UNAUTHORIZED = '0';
```
lukehesluke, line 18: Env var values must be strings. This is also what the node.js code base expects (https://github.com/nodejs/node/search?q=NODE_TLS_REJECT_UNAUTHORIZED&unscoped_q=NODE_TLS_REJECT_UNAUTHORIZED)

PR #147, opened 2020-08-19
feat(amending-order-quote): Add test: c2-with-different-details

lukehesluke:
* Update the existing amend-c1-test to use the same describe-block-for-each-stage-pattern as other tests


packages/openactive-integration-tests/test/features/core/amending-order-quote/implemented/amend-c1-test.js
```diff
@@ -59,26 +60,31 @@ FeatureHelper.describeFeature(module, {
...
-      .beforeSetup()
-      .successChecks()
-      .validationTests();
-    const c1 = (new C1({
-      state, flow, logger,
-    }))
-      .beforeSetup()
-      .successChecks()
-      .validationTests();
+    describe('Get Opportunity Feed Items', () => {
```
lukehesluke, line 63: I've updated this existing test to use the same describe-block-for-each-stage-pattern as other tests.

I assume that this pattern exists in order to keep the test output consistent

I initially didn't want to do this because I thought it might be possible to only use describe() blocks to describe tests that could feasibly be run in parallel. I belive this is usually best practice (so also thinks this person: https://stackoverflow.com/a/39856222) as it allows you to turn up test parallelization if tests become slow and will also allow porting to other test frameworks that do not have synchronous guarantuees about describe blocks.

However, I realised that that would not really be possible without changing the architecture quite a lot (e.g. the functions that create describe() / it() / afterAll() / etc blocks would all need to work differently). And changing the architecture would mean that the test output would not be anywhere near as consistent (not without a lot of effort anyway), so I've decided to go with the architecture rather than against.

Maybe we can think about the parallelization question later. Much later after core tests have been implemented :)

packages/openactive-integration-tests/test/features/core/amending-order-quote/implemented/c2-with-different-details-test.js
```diff
@@ -0,0 +1,121 @@
...
+        .successChecks()
+        .validationTests();
+    });
+  });
+
+  /** Fetch some new opportunities, amend the existing order with a C2 request, and then complete it */
+  describe('Second Attempt - C2 -> B', () => {
+    const { state, flow } = getOpportunityFeedItemsWithNewState();
+
+    describe('C2', () => {
```
nickevansuk, line 96: Just checking (not sure if we can easily test this? Perhaps with request logging?) that the Second Attempt does not call C1 within the second flow, and is happy to jump straight to C2 and ignore C1?

We mentioned previously that C1 might be automatically called if it was not run within a particular flow?

nickevansuk (approved):
Code looks good - one thing to test / confirm?

packages/openactive-integration-tests/test/features/core/amending-order-quote/implemented/c2-with-different-details-test.js
lukehesluke: Good point, this will call C1. That's a separate test. I'll update this one

packages/openactive-integration-tests/test/helpers/flow-helper.js
```diff
@@ -13,9 +13,19 @@ const pMemoize = require("p-memoize");
...
+   *   ensure that the previous stage is called if it hasn't already. e.g.
+   *   C2 will call C1 if it hasn't already been called.
+   *
+   *   Set this to false in order to test unconventional flows like calling
+   *   C2 without calling C1 before.
+   *
+   *   Defaults to true.
    */
-  constructor (helper) {
+  constructor (helper, { doRunPreviousStages = true } = {}) {
```
lukehesluke, line 26: I almost wonder if we want to permanently turn this off 🤔

packages/openactive-integration-tests/test/features/core/amending-order-quote/implemented/amend-c1-test.js
lukehesluke: Right. That's helpful to know. Thanks 🐩

packages/openactive-integration-tests/test/features/core/amending-order-quote/implemented/c2-with-different-details-test.js
lukehesluke: With the latest change, I checked the test log and there was only one C1 request logged (https://openactive-reference-implementation.azurewebsites.net/api/openbooking/order-quote-template/{uuid})

packages/openactive-integration-tests/test/helpers/flow-helper.js
lukehesluke: But I'm disinclined to rock the boat at the moment

packages/openactive-integration-tests/test/helpers/flow-helper.js
```diff
@@ -27,38 +37,48 @@ class FlowHelper {
   }, { cachePromiseRejection: true });
 
   C1 = pMemoize(async () => {
-    await this.getMatch();
-    if (!this.state.getMatchResponseSucceeded) throw Error('Pre-requisite step failed: opportunity feed extract failed');
+    if (this._doRunPreviousStages) {
+      await this.getMatch();
```
nickevansuk, line 41: Interesting... I'd understood that this was needed to ensure the stages were run in sequence? Otherwise the describe blockers weren't guaranteed to run in parallel?

Does this not force each describe block to "await" its dependencies, up the chain?

nickevansuk (commented):
I realise I might have missed this in your previous comment as you were saying something similar to this...

It looks like Jest does run describe locks in series, but should we rely on this?
- https://stackoverflow.com/questions/39831961/are-tests-inside-one-file-run-in-parallel-in-jest#comment84720704_39856222
- https://stackoverflow.com/a/61386996/590558

packages/openactive-integration-tests/test/features/core/amending-order-quote/implemented/c2-with-different-details-test.js
```diff
@@ -91,7 +90,14 @@ FeatureHelper.describeFeature(module, {
 
   /** Fetch some new opportunities, amend the existing order with a C2 request, and then complete it */
   describe('Second Attempt - C2 -> B', () => {
-    const { state, flow } = getOpportunityFeedItemsWithNewState();
+    const state = new RequestState(logger, {
+      uuid,
+    });
+    const flow = new FlowHelper(state, {
+      stagesToSkip: new Set(['C1']),
```
nickevansuk, line 97: Interesting use of `Set()` here - is this pattern normal for config (not seen a library ask for this kind of thing in `options` type setups before?)

nickevansuk (approved):
One question, otherwise looks good!

packages/openactive-integration-tests/test/features/core/amending-order-quote/implemented/c2-with-different-details-test.js
nickevansuk: Would it not be better to move the Set() into the constructor of FlowHelper?

packages/openactive-integration-tests/test/helpers/flow-helper.js
```diff
@@ -13,52 +17,86 @@ const pMemoize = require("p-memoize");
...
+   * @param {Set<StageIdentifier>} [options.stagesToSkip] Any stages included
+   *   in this set will be skipped. e.g. with stagesToSkip = new Set(['c2']),
+   *   the C2 stage will be skipped.
+   *
+   *   This can be used to test unconventional flows
    */
-  constructor (helper) {
+  constructor (helper, { stagesToSkip = new Set() } = {}) {
     this.state = helper;
+    this._stagesToSkip = stagesToSkip
```
nickevansuk, line 29:
```suggestion
    this._stagesToSkip = new Set(stagesToSkip);
```

nickevansuk, line 29: Something like this...

```suggestion
  constructor (helper, { stagesToSkip = [] } = {}) {
    this.state = helper;
    this._stagesToSkip = new Set(stagesToSkip);
```

PR #155, opened 2020-08-24
chore(integration-tests): Increase TypeScript coverage to all but problem areas

lukehesluke:


packages/openactive-integration-tests/test/helpers/feature-helper.js
```diff
@@ -11,7 +11,7 @@ const IMPLEMENTED_FEATURES = global.IMPLEMENTED_FEATURES;
 /**
  * @typedef {import('../types/OpportunityCriteria').OpportunityCriteria} OpportunityCriteria
  *
- * @typedef {(opportunityType: string) => [OpportunityCriteria]} CreateSingleOportunityCriteriaTemplateFn
+ * @typedef {(opportunityType: string) => OpportunityCriteria[]} CreateSingleOportunityCriteriaTemplateFn
```
lukehesluke, line 14: One invokee of the feature helper uses a Single opportunity criteria template function which returns 2 criteria rather than just one. This seems odd but doesn't break anything, so I've updated this to allow a variably sized array

PR #159, opened 2020-09-02
feat(common-error-conditions): Create not-found-error test

### LINKED ISSUE #157: common-error-conditions - not-found - Expect a NotFoundError for Orders that do not exist
**Feature**

**Use Case**

Runs Order Cancellation for an non-existent Order (with a fictional UUID), expecting an NotFoundError error to be returned

In the Booking API spec: https://openactive.io/open-booking-api/EditorsDraft/#order-deletion

Note: a NotFoundError can also be returned for the Order Status endpoint. This is not being tested here, because the Order Status endpoint is optional and only required endpoints are tested by core tests

**Why this hasn't been covered by existing tests**

There are no existing tests which test Delete Order - let alone for a non-existent Order

lukehesluke:
Closes #157 

## QA

Screenshot 2020-09-02 at 11 29 44
Screenshot 2020-09-02 at 11 30 06
Screenshot 2020-09-02 at 11 30 28


packages/openactive-integration-tests/test/features/.eslintrc.json
```diff
@@ -9,8 +9,8 @@
   },
   "rules": {
     "object-curly-newline": ["error", {
-        "ObjectExpression": "always",
-        "ObjectPattern": { "multiline": true },
```
lukehesluke, line 13: This lint rule was making it so that even `{}` or `{ x }` required new lines. I changed it to `minProperties: 3` to match the export declaration. Let me know if there are any problems with this

packages/openactive-integration-tests/test/features/core/availability-check/feature.json
```diff
@@ -6,7 +6,7 @@
   "explainer": "",
   "specificationReference": "https://www.openactive.io/open-booking-api/EditorsDraft/#step-by-step-process-description-0",
   "required": true,
-  "coverageStatus": "partial",
+  "coverageStatus": "complete",
```
lukehesluke, line 9: This is complete now according to the spreadsheet

package.json
```diff
@@ -12,7 +12,7 @@
     "install-broker": "cd packages/openactive-broker-microservice && npm install",
     "install-interface-criteria": "cd packages/test-interface-criteria && npm install",
     "install-tests": "cd packages/openactive-integration-tests && npm install",
-    "install": "npm run install-interface-criteria && npm run gen-interface-criteria-types && npm run install-broker && npm run install-tests",
+    "install": "npm run install-interface-criteria && npm run install-broker && npm run install-tests",
```
lukehesluke, line 15: This has now been fixed as per your comment in a previous PR

packages/openactive-integration-tests/test/features/core/common-error-conditions/implemented/not-found-test.js
```diff
@@ -0,0 +1,49 @@
...
+      const normalizedKey = (key === 'id' || key === 'type')
+        ? `@${key}`
+        : key;
+      const normalizedValue = normalizeAtFields(value);
+      newObject[normalizedKey] = normalizedValue;
+    }
+    return newObject;
+  }
+  return oaObject;
+}
```
lukehesluke, line 28: Of course, this function should be moved out of this file when it's used by other tests. Let me know if this sort of function already existed. I wasn't able to find it

packages/openactive-integration-tests/test/features/.eslintrc.json
lukehesluke: It seems like we should have a consistent rule for all of these. But I've opted out of that for now as that would require making lots of (auto-fixed) changes to the codebase than would not be reasonable for this PR

nickevansuk: This is great, I was wondering about this! Agreed

packages/openactive-integration-tests/test/features/core/common-error-conditions/implemented/not-found-test.js
nickevansuk: Actually the booking spec forces @type - so we should only accept @type and @id within the requests and responses (ref https://github.com/openactive/open-booking-api/issues/141)

nickevansuk (changes requested):
Looks great, one minor thing

nickevansuk (approved):
Super!

PR #161, opened 2020-09-02
fix: Fix request-helper._request function

lukehesluke:
This was previously not passing params properly for get requests. It also had confusingly named arguments

### DIFF
--- a/packages/openactive-integration-tests/test/helpers/request-helper.js
+++ b/packages/openactive-integration-tests/test/helpers/request-helper.js
@@ -69,7 +69,6 @@ class RequestHelper {
    */
   async get(stage, url, requestOptions) {
     return await this._request(stage, 'GET', url, null, requestOptions);
-    // return await this._request2(stage, )
   }
 
   /**
--- a/packages/openactive-integration-tests/test/helpers/request-helper.js
+++ b/packages/openactive-integration-tests/test/helpers/request-helper.js
@@ -9,6 +9,12 @@ const c2req = require("../templates/c2-req.js");
 const { bReqTemplates } = require("../templates/b-req.js");
 const ureq = require("../templates/u-req.js");
 
+/**
+ * @typedef {import('./logger').BaseLoggerType} BaseLoggerType
+ * @typedef {import('chakram').RequestMethod} RequestMethod
+ * @typedef {import('chakram').RequestOptions} RequestOptions
+ */
+
 const REQUEST_HEADERS = config.get("sellers.primary.requestHeaders");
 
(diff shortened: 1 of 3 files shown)


packages/openactive-integration-tests/test/helpers/request-helper.js
```diff
@@ -41,23 +62,52 @@ class RequestHelper {
     return response;
   }
 
-  async get(stage, url, headers) {
-    return await this._request(stage, 'GET', url, null, headers);
```
lukehesluke, line 45: This would call chakram.get(url, null, headers). If you look at the docs (http://dareid.github.io/chakram/jsdoc/module-chakram.html#.get), you can see that chakram.get only takes 2 args - therefore, "headers" was being ignored

packages/openactive-integration-tests/test/helpers/request-helper.js
```diff
@@ -9,30 +9,51 @@ const c2req = require("../templates/c2-req.js");
...
 
 class RequestHelper {
+  /**
+   * @param {BaseLoggerType} logger
+   */
   constructor(logger) {
     this.logger = logger;
   }
 
-  async _request(stage, method, url, params, headers) {
```
lukehesluke, line 23: "params" was actually the JSON body.

"headers" was actually params, which can include headers (e.g. `{ timeout: 1000, headers: {} }`).

So I've renamed these to be more clear and to use chakram.request, which has a fixed function signature, rather than chakram.get/chakram.post/etc

packages/openactive-integration-tests/test/types/chakram/index.d.ts
```diff
@@ -22,9 +22,13 @@ declare module 'chakram' {
     responseTime: number;
   }
 
-  type RequestOptions = Pick<Options, 'headers' | 'body'>;
+  // `url` is required in request's options, but is not needed in chakram,
+  // because `url` is provided as a separate argument to params
+  export type RequestOptions = Omit<Options, 'url'>;
```
lukehesluke, line 27: Yes, chakram's "params" arg has the same type as the `request` library. They state so here: http://dareid.github.io/chakram/jsdoc/module-chakram.html#.request. Though with the apparent exception that `url` is missing. I expect there are more exceptions tbh but nothing's broken so far

packages/openactive-integration-tests/test/helpers/request-helper.js
nickevansuk: Good catch, that was quite a hack previously!

nickevansuk (approved):
Great improvement!

PR #169, opened 2020-09-07
chore(integration-tests): Expand ESLint coverage

lukehesluke:


packages/openactive-integration-tests/package.json
```diff
@@ -7,8 +7,8 @@
     "start": "jest",
     "test": "npm run lint && tsc && npm run doc-gen",
     "debug": "node --inspect ./node_modules/jest/bin/jest.js --runInBand",
-    "lint": "eslint \"test/features/**/*.js\"",
-    "lint-fix": "eslint \"test/features/**/*.js\" --fix",
+    "lint": "eslint \"test/**/*.js\"",
```
lukehesluke, line 10: Now, if any new files are added anywhere in the code, they will be linted by default.

It's only the files that have un-auto-fixable errors that are explicitly ignored by lint (for now)

PR #170, opened 2020-09-07
feat(common-error-conditions): Add incomplete-customer-details test

### LINKED ISSUE #165: common-error-conditions - incomplete-customer-details - Expect an IncompleteCustomerDetailsError when customer details are incomplete
**Feature**

**Use Case**

Run each of C2 and B for a valid opportunity, with customer details incomplete, expecting an IncompleteCustomerDetailsError to be returned.

(C1 is ignored because customer details are not accepted for C1)

In the Booking API spec: https://openactive.io/open-booking-api/EditorsDraft/#order-creation-orderquote-orderproposal-and-order-error-responses

lukehesluke:
closes #165


.vscode/launch.json
```diff
@@ -5,30 +5,14 @@
     "version": "0.2.0",
     "configurations": [
         {
-            "name": "integration-tests - core/availability-check/opportunity-full",
+            "name": "integration-tests - core/common-error-conditions/incomplete-customer-details",
```
lukehesluke, line 8: Constantly editing this configuration is not ideal. Ideally, we'd have a "Integration Test this file", which is possible with `${file}` (https://github.com/microsoft/vscode-recipes/tree/master/debugging-mocha-tests#configure-launchjson-file-for-your-test-framework) but that wouldn't work here as it'd give the full path including `packages/openactive-integration-tests`.

Not very important though

packages/openactive-integration-tests/test/features/core/common-error-conditions/implemented/incomplete-customer-details-test.js
```diff
@@ -0,0 +1,112 @@
...
+   * Fetch some opportunities and run C1
+   *
+   * Note: This generates jest blocks like `beforeAll()`, `it()`, etc. Therefore, this must be run within a `describe()` block
+   *
+   * @param {RequestState} state
+   * @param {FlowHelper} state
+   */
+  function runC1(state, flow) {
+    beforeAll(async () => {
+      await state.fetchOpportunities(orderItemCriteria);
```
lukehesluke, line 34: I've experimented by not including the `await chakram.wait()` line, which I don't think is needed as we await chakram requests (as far as I'm aware) as soon as they happen. Seems to work 👍

.vscode/launch.json
nickevansuk: I think Jest can handle the full path?

packages/openactive-integration-tests/test/features/core/common-error-conditions/implemented/incomplete-customer-details-test.js
nickevansuk: Ye think this works if only waiting for one request...

PR #178, opened 2020-09-10
Coverage/common error conditions  unknown order

civsiv:


lukehesluke (changes requested):

nickevansuk (approved):
Approve based on @lukehesluke comments being addressed

PR #183, opened 2020-09-15
Feature/common error conditions  order already exists

### LINKED ISSUE #167: common-error-conditions - order-already-exists - Expect an OrderAlreadyExistsError if an Order UUID exists but with different OrderItems
**Feature**

**Use Case**

Do a successful C1, C2, B run. Then, run B again for the same Order UUID, but with different OrderItems. Expect an OrderAlreadyExistsError.

In the Open Booking API spec: https://openactive.io/open-booking-api/EditorsDraft/#order-creation-orderquote-orderproposal-and-order-error-responses

civsiv:
closes #167


packages/openactive-integration-tests/test/shared-behaviours/b.js
```diff
@@ -2,6 +2,12 @@ const { expect } = require('chakram');
 const sharedValidationTests = require('./validation');
 
 class B {
+  /**
+   * @param {object} args
+   * @param {InstanceType<import('../helpers/request-state')['RequestState']>} args.state
+   * @param {import('../helpers/flow-helper').FlowHelperType} args.flow
+   * @param {import('../helpers/logger').BaseLoggerType} args.logger
+   */
```
lukehesluke, line 10: yasss

packages/openactive-integration-tests/test/features/core/common-error-conditions/implemented/order-already-exists-test.js
```diff
@@ -0,0 +1,84 @@
...
+        state, flow, logger,
+      }))
+        .beforeSetup()
+        .successChecks()
+        .validationTests();
+    });
+
+    it('should get a new set of OrderItems and try booking again, but with the same UUID', async () => {
+      // The following requests are not using the helper functions that are used above. This is because
+      // those requests are cached (using pMemoize) and therefore would not go back to the Booking System.
```
lukehesluke, line 73: @civsiv I think that these should use the helper functions because they provide useful stuff like consistent test reporting messages (and structure) as well as performing validation tests (which should always be performed on responses

If you look at core/amending-order-quote/implemented/amend-c1-test.js, there is an approach there that allows it. UUID can be shared between states by including it in new RequestState

lukehesluke (changes requested):

.vscode/launch.json
```diff
@@ -5,21 +5,24 @@
...
-            "name": "integration-tests - core/common-error-conditions/incomplete-customer-details",
+            "name": "integration-tests - core/common-error-conditions/order-already-exists",
             "request": "launch",
             "runtimeArgs": [
                 "run-script",
                 "start-tests",
                 "--",
                 "--runInBand",
-                "test/features/core/common-error-conditions/implemented/incomplete-customer-details-test.js"
+                "test/features/core/common-error-conditions/implemented/order-already-exists-test.js"
```
nickevansuk, line 15: We should fix this https://github.com/openactive/openactive-test-suite/issues/184 next time we change this, to stop us updating this in each PR

nickevansuk (approved):
Assuming @lukehesluke's comments are addressed

packages/openactive-integration-tests/test/helpers/request-state.js
```diff
@@ -207,14 +207,33 @@ class RequestState {
     return isResponse(this.c1Response);
   }
 
+  /**
+   * As leasing is optional, getting the totalPaymentDue should not be dependent on C1 or C2 being completed.
```
nickevansuk, line 211: Although Leasing is optional C1 and C2 are not (ref: https://openactive.io/open-booking-api/EditorsDraft/#broker-contract-with-booking-system)

So `totalPaymentDue` from the response must always be used. See also https://github.com/openactive/open-booking-api/issues/152

nickevansuk (changes requested):

packages/openactive-integration-tests/test/helpers/request-state.js
civsiv: Ah ok, in that case the RefImpl needs to be changed as it returns a PaymentMismatchError when it shouldn't

PR #188, opened 2020-09-23
Fixing Facility Use template

NemanjaStefanovic:
using provider instead of organizer in Facility Use template


nickevansuk (changes requested):
Looks good, just need to fix the lint issues so CI passes

PR #191, opened 2020-09-25
feat: Improve robustness of coverage for FacilityUse

NemanjaStefanovic:
Fix issues with tests relating to FacilityUse

Also fixes issue with request body output that was introduced in #161


nickevansuk (approved):
Looks great!

PR #193, opened 2020-09-28
feat(minimal-proposal): Add successful-book test

nickevansuk:


packages/openactive-integration-tests/test/helpers/flow-helper.js
```diff
@@ -79,9 +87,40 @@ class FlowHelper {
...
+    this._assertStageShouldNotBeSkipped('getFeedUpdateAfterP');
+    if (!this._stagesToSkip.has('P')) {
+      await this.P();
+      if (!this.state.PResponseSucceeded) throw Error('Pre-requisite step failed: P failed');
+    }
+
+    return await this.getOrderAfterPPromise;
+  }, { cachePromiseRejection: true });
+
+  BAfterP = pMemoize(async () => {
```
lukehesluke, line 112: @nickevansuk I think we could do with rethinking the flow-helper logic so that you define specific flows. e.g. we can have the default flow be `C1 -> C2 -> B`, as well as a cancellation flow which is `(default flow) -> U -> getFeedUpdate`

This way, we don't have to have things like `BAfterP` and `getFeedUpdateAfterU` and the `_stagesToSkip`. It would require a bit of work though e.g.  it might be helpful to have the "flow item" store its own state rather than request-state just having a `BResponse`.. 🤔

lukehesluke: This would make it easier to slot actions in also. I wrote the generic `awaitStage` function out of the dread of thinking of having to write a custom bit of flow logic for each action's use case in its given flow but it could cause issues in other kinds of flows

packages/openactive-integration-tests/test/helpers/flow-helper.js
```diff
@@ -79,9 +87,40 @@ class FlowHelper {
     return this.state.putOrder();
   }, { cachePromiseRejection: true });
 
+  P = pMemoize(async () => {
+    this._assertStageShouldNotBeSkipped('P');
+    this.getOrderAfterPPromise = this.state.getOrderAfterP();
```
nickevansuk, line 92: This also makes the assumption that P is always followed by a `getOrder` check, which should not be the case as P does not add anything to the Orders Feed ("A" does, but which is simulated by the test interface action)

packages/test-interface-criteria/src/criteria/sharedConstraints.js
```diff
@@ -0,0 +1,13 @@
+/**
+ * @typedef {import('../types/Criteria').OfferConstraint} OfferConstraint
+ */
+
+/** @type {OfferConstraint} */
+function supportsMinimalProposalFlow(offer) {
+  return Array.isArray(offer.openBookingFlowRequirement)
+    && offer.openBookingFlowRequirement.includes('https://openactive.io/OpenBookingApproval');
```
nickevansuk, line 8: This should be _only_ `https://openactive.io/OpenBookingApproval`, as the test doesn't support attendee details capture or other `openBookingFlowRequirement`s

packages/openactive-integration-tests/test/helpers/flow-helper.js
nickevansuk: Is this primarily about figuring out which flow stage waits on what? E.g. that in some cases B depends on C2,  getFeedUpdate depends on U, etc?

I wonder given the number of potential variants mean we should have defaults and "custom", where custom is declares at the top of the test, perhaps as a simple array of stages? We could then define some constants which are just pre-defined arrays e.g. `['C1', 'C2', 'B']`, `['C1', 'C2', 'B', `U`, `getFeedUpdate`]` ?

Perhaps B could then look within the array to figure out what's before it that it should be waiting on?

We also need some way of it knowing that any "get from feed" type of stage need to start before the previous stage (to collect whatever result from it), perhaps we just do that by convention? So `get*` will always start before the previous stage?)

nickevansuk: Also some tests are `['GetMatch', 'C1', 'C2', 'B', 'GetMatch']`

packages/test-interface-criteria/src/criteria/sharedConstraints.js
nickevansuk: Have resolved this

nickevansuk:
Closing noting outstanding refactor

PR #216, opened 2020-10-08
Add docker support

nathansalter:
This adds experimental Docker support so you can run the tests without needing a specific version of npm installed on your machine. Also allows for a better CI experience as this can be added in the CI pipeline.

In future we should work on reducing the size of the containers (currently ~1GB each) and providing them tagged in the public docker repositories.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,19 @@
+FROM node:12.19.0-alpine3.12 AS openactive.integration_test
+
+WORKDIR /openactive
+
+COPY packages /openactive/packages
+COPY package.json /openactive/package.json
+COPY package-lock.json /openactive/package-lock.json
+COPY start.js /openactive/start.js
+
+WORKDIR /openactive
+RUN npm install
+
+RUN cp -rf /openactive/packages/test-interface-criteria /openactive/packages/openactive-broker-microservice/node_modules/
+RUN cp -rf /openactive/packages/test-interface-criteria /openactive/packages/openactive-integration-tests/node_modules/
+
+VOLUME /openactive/packages/openactive-broker-microservice/config
+VOLUME /openactive/packages/openactive-integration-tests/config
+
+ENTRYPOINT npm start
--- a/README.md
+++ b/README.md
@@ -33,6 +33,19 @@ A file named `{NODE_ENV}.json` can also be placed alongside the `default.json`, 
 
 For more information see the [documentation](https://github.com/lorenwest/node-config/wiki/Environment-Variables#node_env).
 
+## Docker
+
+You can run these files on docker by first building the containers:
+```bash
+docker build -t openactive:latest ./
+```
+
+Then simply run the tests:
+
+```
+docker run --rm -v $PWD/packages/openactive-broker-microservice/config:/openactive/packages/openactive-broker-microservice/config openactive
+```
+
 ## Installation
 ```bash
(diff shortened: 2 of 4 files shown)


README.md
```diff
@@ -33,6 +33,29 @@ A file named `{NODE_ENV}.json` can also be placed alongside the `default.json`,
 
 For more information see the [documentation](https://github.com/lorenwest/node-config/wiki/Environment-Variables#node_env).
 
+## Docker
+
+You can run these files on docker by first building the containers:
+```bash
+cd packages
+docker build -f DockerMicroservice -t openactive.microservice:latest ./
+docker build -f DockerIntegration -t openactive.integration:latest ./
```
nickevansuk, line 42: Just looking at this, is there any reason why we couldn't use a single dockerfile?

`npm run start-broker` and `npm run start-tests` can run in the root directory?

Also `npm start` runs both together and kills the mircoservice upon test completion... so this would allow users of the dockerfile to have this ability

It would also remove the need to copy test-interface-criteria, and means that `npm install` in the root (which installs all packages) would also work

nickevansuk (commented):
@nathansalter what do you think?

README.md
nathansalter: Took me ages to work out why I couldn't get this to work, turns out it was a node versioning issue. I've also moved to alpine containers so we should be able to look into publishing a container on docker hub so people can just do `docker run --rm openactive.test-suite` in future

nickevansuk: Great idea!

packages/DockerMicroservice
```diff
@@ -0,0 +1,15 @@
+FROM node AS openactive.microservice
```
nickevansuk, line 1: Noticing this file is still here. Should this have been deleted in this last commit?

Dockerfile
```diff
@@ -0,0 +1,19 @@
...
+WORKDIR /openactive
+RUN npm install
+
+RUN cp -rf /openactive/packages/test-interface-criteria /openactive/packages/openactive-broker-microservice/node_modules/
+RUN cp -rf /openactive/packages/test-interface-criteria /openactive/packages/openactive-integration-tests/node_modules/
+
+VOLUME /openactive/packages/openactive-broker-microservice/config
+VOLUME /openactive/packages/openactive-integration-tests/config
+
+ENTRYPOINT npm start
```
nickevansuk, line 19: should we be allowing args to be passed through (e.g. [here](https://stackoverflow.com/questions/32727594/how-to-pass-arguments-to-shell-script-through-docker-run))? Are they already allowed by this?

Dockerfile
```diff
@@ -0,0 +1,19 @@
...
+
+WORKDIR /openactive
+RUN npm install
+
+RUN cp -rf /openactive/packages/test-interface-criteria /openactive/packages/openactive-broker-microservice/node_modules/
+RUN cp -rf /openactive/packages/test-interface-criteria /openactive/packages/openactive-integration-tests/node_modules/
+
+VOLUME /openactive/packages/openactive-broker-microservice/config
+VOLUME /openactive/packages/openactive-integration-tests/config
+
```
nickevansuk, line 18: Perhaps also worth `EXPOSE 3000` here to allow the microservice status page to be viewed?

nickevansuk (commented):
Further comments

Issue #225, opened 2020-10-12
agent-broker - customer-included - B request should succeed if customer is included in Order in AgentBroker mode.


NemanjaStefanovic:
**Feature**

agent-broker
Request should succeed if Customer is included in Order.

**Use Case**

Sending B request with Customer field. Request should succeed.

[Open Booking API spec](https://openactive.io/open-booking-api/EditorsDraft/#agentbroker).

PR #227, opened 2020-10-12
Feature/customer included agent broker

NemanjaStefanovic:
Second test in Row 10 (https://docs.google.com/spreadsheets/d/1Wny82fY-oU45beN6u55X9CMzdpf-j242fp3RG8_WJ3E/edit?pli=1#gid=1383644278)

Issue #225

Note:
possible duplication of: test\features\payment\simple-book-with-payment\implemented\with-payment-property-test.js
(simple successful request, with default values)


lukehesluke (changes requested):
@NemanjaStefanovic I think it would be good to explicitly set the AgentBroker. The way it's currently written only works because the default c1, c2, etc templates have `brokerRole` set to `AgentBroker`. But if these defaults were changed, these tests would no longer be accurate.

You could use an optional `brokerRole` arg in the RequestState constructor (which can just default to AgentBroker)

lukehesluke (approved):

Issue #240, opened 2020-10-14
seller-requested-cancellation-message - seller-requested-cancellation-message - C1, C2, B, Action (test:SellerRequestedCancellationWithMessageSimulateAction), and Orders feed should include successful cancellation, including a cancellationMessage


NemanjaStefanovic:
**Feature**

cancellation (by seller)

**Use Case**

C1, C2, B, Action (test:SellerRequestedCancellationWithMessageSimulateAction), and Orders feed should include successful cancellation, including a cancellationMessage

[Open Booking API spec](https://www.openactive.io/open-booking-api/EditorsDraft/#seller-requested-cancellation).

PR #243, opened 2020-10-14
Coverage/seller requested cancellation message

NemanjaStefanovic:
**Draft PR**

Added test for seller cancellation with message

Depends on #215 
Issue #240


nickevansuk:
Merged with issue raised #264

Issue #236, opened 2020-10-14
Tests should run for single occupancy booking systems


nickevansuk:
Tests currently fail if maximumAttendeeCapacity is a maximum of 1.

The ref impl should be updated to exhibit this behaviour, for "IndividualFacilityUse" only.

Issue #237, opened 2020-10-14
Base URL should not require a trailing '/'


nickevansuk:
Test suite and ref impl should be updated as per https://github.com/openactive/open-booking-api/issues/161

nickevansuk:
This has now been resolved by https://github.com/openactive/openactive-test-suite/pull/246

PR #246, opened 2020-10-14
fix: Capacity limit for IndividualFacilityUse and use of Harvest Start Time

nickevansuk:
Fixes these issues:
- `IndividualFacilityUse` is not allowed `Slot` with `remainingUses` of greater than one, which causes issues with "multiple" opportunity tests (https://github.com/openactive/openactive-test-suite/issues/236)
- Additionally the current time is referenced in test-interface-criteria which makes the bucket allocation inconsistent between harvesting and the test run.

### DIFF
--- a/packages/openactive-broker-microservice/app.js
+++ b/packages/openactive-broker-microservice/app.js
@@ -306,7 +306,7 @@ app.get('/health-check', function (req, res) {
   }
 });
 
-// Config endpoint used to set global variables within the integration tests
+// Config endpoint used to get global variables within the integration tests
 app.get('/config', function (req, res) {
   res.json({
     // Allow a consistent startDate to be used when calling test-interface-criteria
--- a/packages/openactive-broker-microservice/app.js
+++ b/packages/openactive-broker-microservice/app.js
@@ -260,9 +260,9 @@ function setFeedIsUpToDate(feedIdentifier) {
         const { childOrphans, totalChildren, percentageChildOrphans } = getOrphanStats();
 
         if (totalChildren === 0) {
-          logError('\nFATAL ERROR: Zero opportunities could be harvested from the opportunities feed.');
-          logError('Please ensure that the opportunities feed conforms to RPDE using https://validator.openactive.io/rpde.\n');
-          throw new Error('Zero opportunities could be harvested from the opportunities feed');
+          logError('\nFATAL ERROR: Zero opportunities could be harvested from the opportunities feeds.');
+          logError('Please ensure that the opportunities feeds conforms to RPDE using https://validator.openactive.io/rpde.\n');
+          throw new Error('Zero opportunities could be harvested from the opportunities feeds');
         } else if (childOrphans === totalChildren) {
           logError(`\nFATAL ERROR: 100% of the ${totalChildren} harvested opportunities do not have a matching parent item from the parent feed, so all integration tests will fail.`);
           logError('Please ensure that the value of the `subEvent` or `facilityUse` property in each opportunity exactly matches an `@id` from the parent feed.\n');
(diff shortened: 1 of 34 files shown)


packages/openactive-broker-microservice/app.js
```diff
@@ -304,8 +306,18 @@ app.get('/health-check', function (req, res) {
   }
 });
 
+// Config endpoint used to set global variables within the integration tests
```
lukehesluke, line 309:
```suggestion
// Config endpoint used to get global variables within the integration tests
```

packages/openactive-integration-tests/test/testEnvironment.js
```diff
@@ -17,9 +17,13 @@ class TestEnvironment extends NodeEnvironment {
     await super.setup();
 
     // Get endpoint URL from Dataset Site
```
lukehesluke, line 19: this comment is now slightly out of date

packages/test-interface-criteria/src/criteria/internal/InternalTestOpportunityBookable.js
```diff
@@ -12,23 +12,16 @@ const { getRemainingCapacity, createCriteria, mustBeWithinBookingWindow } = requ
  * @type {OpportunityConstraint}
  */
 function remainingCapacityMustBeAtLeastTwo(opportunity) {
-  return getRemainingCapacity(opportunity) > 1;
+  return getRemainingCapacity(opportunity) > (hasCapacityLimitOfOne(opportunity) ? 0 : 1);
 }
```
lukehesluke, line 16: Maybe worth adding a comment that 2 is needed in other cases because the multiple OrderItem tests use 2 of the same item.

I feel that this function and the opportunityReuseKey logic should be derived from the same foundation in some way (e.g. a config?). Otherwise, it seems like it'd be easy to change one without realising the other needs changing (e.g. "maybe we should have 3 items with the same reuse key")

However, that would require some thinking time and I understand this is urgent

packages/openactive-integration-tests/test/helpers/request-state.js
```diff
@@ -87,7 +88,7 @@ class RequestState {
      */
     this.testInterfaceResponses = await Promise.all((this.orderItemCriteriaList || []).map(async (orderItemCriteriaItem, i) => {
       // If an opportunity is available for reuse, return it
-      if (orderItemCriteriaItem.hasOwnProperty('opportunityReuseKey') && reusableOpportunityPromises.has(orderItemCriteriaItem.opportunityReuseKey)) {
+      if (orderItemCriteriaItem.hasOwnProperty('opportunityReuseKey') && orderItemCriteriaItem.opportunityReuseKey !== null && reusableOpportunityPromises.has(orderItemCriteriaItem.opportunityReuseKey)) {
```
lukehesluke, line 91:
```suggestion
      if (!_.isNil(orderItemCriteriaItem.opportunityReuseKey') && reusableOpportunityPromises.has(orderItemCriteriaItem.opportunityReuseKey)) {
```

and lodash would need to be imported at the top

This is a bit more generic ("field not present", "field value is null or undefined" would all be caught by this) and doesn't use the `{}.hasOwnProperty()` method, which [can cause issues](https://eslint.org/docs/rules/no-prototype-builtins)

packages/openactive-integration-tests/test/helpers/request-state.js
```diff
@@ -98,17 +99,17 @@ class RequestState {
         : this.requestHelper.createOpportunity(orderItemCriteriaItem.opportunityType, orderItemCriteriaItem.opportunityCriteria, i, seller['@id'], seller['@type']);
 
       // If this opportunity can be reused, store it
-      if (orderItemCriteriaItem.hasOwnProperty('opportunityReuseKey')) {
+      if (orderItemCriteriaItem.hasOwnProperty('opportunityReuseKey') && orderItemCriteriaItem.opportunityReuseKey !== null) {
```
lukehesluke, line 102:
```suggestion
      if (!_.isNil(orderItemCriteriaItem.opportunityReuseKey)) {
```

lukehesluke (changes requested):

packages/test-interface-criteria/src/criteria/internal/InternalTestOpportunityBookable.js
nickevansuk: Yes great point, have added an issue for this to be picked up in future: https://github.com/openactive/openactive-test-suite/issues/247

lukehesluke (approved):

nickevansuk:
Also added a fix for https://github.com/openactive/openactive-test-suite/issues/237 to this PR, for expediency

Issue #251, opened 2020-10-15
access-code - manual-access-codes - Receiving access code when calling B endpoint successfully


NemanjaStefanovic:
**Feature**

access-code

**Use Case**

Receiving access code when calling B endpoint successfully

[Open Booking API spec](https://www.openactive.io/open-booking-api/EditorsDraft/#text-based-access-control).

PR #252, opened 2020-10-15
Coverage/access code manual

NemanjaStefanovic:
@nickevansuk  I am checking only on B endpoint as C1 and C2 shouldn't contain access code, 
but https://docs.google.com/spreadsheets/d/1Wny82fY-oU45beN6u55X9CMzdpf-j242fp3RG8_WJ3E/edit?pli=1#gid=1383644278 suggests that it should also check C1 and C2 which doesn't make sense (we should receive access code only if we successfully book right? can you just confirm that?)

Issue #251


packages/openactive-integration-tests/test/features/access/access-code/implemented/manual-access-codes-test.js
```diff
@@ -0,0 +1,72 @@
...
+  describe('B', function () {
+    (new B({
+      state, flow, logger,
+    }))
+      .beforeSetup()
+      .successChecks()
+      .validationTests();
+
+    it('Response should contain Access Code', function () {
+      chai.expect(state.bResponse.body.orderedItem[0].accessCode[0].description).to.equal('1234');
```
nickevansuk, line 69: This should be a schema check rather than a specific value check, as we're a real system could return any valid code.

E.g.
https://github.com/openactive/openactive-test-suite/blob/7e02bc9e17486af14d888267aae0505c15bcfa3d/packages/openactive-integration-tests/test/features/core/dataset-site/implemented/dataset-site-jsonld-valid-test.js#L30-L35

nickevansuk (changes requested):

lukehesluke:
@NemanjaStefanovic 

> I am checking only on B endpoint as C1 and C2 shouldn't contain access code,
but https://docs.google.com/spreadsheets/d/1Wny82fY-oU45beN6u55X9CMzdpf-j242fp3RG8_WJ3E/edit?pli=1#gid=1383644278 suggests that it should also check C1 and C2 which doesn't make sense (we should receive access code only if we successfully book right? can you just confirm that?)

Yes, you're right. The test is written wrong on the spreadsheet. The [spec](https://openactive.io/open-booking-api/EditorsDraft/1.0CR3/#schema-orderitem) specifies that it is not applicable for OrderQuotes (C1 & C2)

packages/openactive-integration-tests/test/features/access/access-code/implemented/manual-access-codes-test.js
```diff
@@ -0,0 +1,77 @@
...
+  describe('B', function () {
+    (new B({
+      state, flow, logger,
+    }))
+      .beforeSetup()
+      .successChecks()
+      .validationTests();
+
+    it('Response should include accessCode with description and name', () => {
+      chakram.expect(state.bResponse).to.have.schema('orderedItem[0].accessCode[0].description', {
```
lukehesluke, line 69: This should check every orderedItem - rather than just the 1st orderedItem

lukehesluke (changes requested):

packages/openactive-integration-tests/test/features/access/access-code/implemented/manual-access-codes-test.js
```diff
@@ -0,0 +1,85 @@
...
+      state.bResponse.body.orderedItem.forEach((orderItem, orderItemIndex) => {
+        chakram.expect(orderItem.accessCode).to.be.an('array');
+
+        orderItem.accessCode.forEach((accessCode, accessCodeIndex) => {
+          chakram.expect(state.bResponse).to.have.schema(`orderedItem[${orderItemIndex}].accessCode[${accessCodeIndex}].name`, {
+            type: 'string',
+          });
+          chakram.expect(state.bResponse).to.have.schema(`orderedItem[${orderItemIndex}].accessCode[${accessCodeIndex}].description`, {
+            type: 'string',
+          });
```
lukehesluke, line 80: Sorry, this is a weird one, but we try to use chai's expect function as much as possible - using chakram's expect' only for .to.have.status(..)`, which is useful

chakram is based off of old chai code, which it added some HTTP request stuff to. Therefore, we prefer chai, which is up-to-date and well documented.

(I personally think we should bin chakram at some point for something like axios. It's made for a pre-promise JS world (though it does have promise support now) and is not typed or well documented)

chai doesn't have `to.have.schema()` but since it's just checking that those values are strings, you can use `be.a('string')` e.g.

```
chai.expect(state.bResponse).to.have.nested.property(`orderedItem[${orderItemIndex}].accessCode[${accessCodeIndex}].name`)
  .that.is.a('string');
```

lukehesluke (changes requested):

packages/openactive-integration-tests/test/features/access/access-code/implemented/manual-access-codes-test.js
NemanjaStefanovic: Ok sure, I didn't know that, 
Nick suggested to use to.have.schema instead of checking the actual value, but your suggestion makes sense, that way we can still check 'structure' without actual values, will change it to chai and that.is.a - string

nickevansuk: Apologies @NemanjaStefanovic for misleading here!

NemanjaStefanovic: np, I learnt something new :)

nickevansuk (approved):

PR #254, opened 2020-10-16
Proposed changes

JoshuaLevett:
Assorted changes made during the usage of the test suite

I've tried to tidy up/clarify what each commit was for (locally it's a bit of a mess!)


nickevansuk:
These fixes have since been incorporated into the test suite - thanks @JoshuaLevett !

PR #257, opened 2020-10-20
Flow Stages - variation

lukehesluke:
**DRAFT - do not review this**


packages/openactive-integration-tests/test/features/approval/minimal-proposal/implemented/accept-proposal-book-test.js
```diff
@@ -42,50 +42,141 @@ FeatureHelper.describeFeature(module, {
...
+  {
+    const { fetchOpportunities, c1A, c1B, c2 } = (new SimpleFlowBuilder({ requestHelper, logger }))
+      .addFetchOpportunities(orderItemCriteriaList)
+      .add('c1', C1FlowStage.create)
+      .add('p', PFlowStage.create) // this will use totalPaymentDue from c1
+      .add('c2', C2FlowStage.create)
+      .add('b', BFlowStage.create, () => ({
+        getInputStateFrom: 'c1', // will use the merged state from fetchOpportunities & c1 - and no further
+      }));
+  }
```
lukehesluke, line 134: some examples

packages/openactive-integration-tests/test/features/approval/minimal-proposal/implemented/accept-proposal-book-test.js
```diff
@@ -42,50 +42,351 @@ FeatureHelper.describeFeature(module, {
...
+        orderItems: fetchOpportunities.getOutput().orderItems,
+        orderProposalVersion: p.getOutput().orderProposalVersion, // (this could also use collectOrderFeedUpdate.getOutput())
+        totalPaymentDue: p.getOutput().totalPaymentDue,
+      }),
+      prerequisite: collectOrderFeedUpdate,
+      logger,
+      requestHelper,
+    });
+  }
+  // ## 1.2. With simple substitutions / helper functions
```
nickevansuk, line 142: Am still skeptical of using merging, even via the helpers, as we could get inconsistent behaviour this way... I wonder if there are other options here?

packages/openactive-integration-tests/test/features/approval/minimal-proposal/implemented/accept-proposal-book-test.js
```diff
@@ -42,50 +42,351 @@ FeatureHelper.describeFeature(module, {
...
+      uuid,
+      sellerId,
+      requestHelper,
+      logger,
+    });
+    const p = PFlowStage.create({
+      getInputState: () => ({
+        orderItems: fetchOpportunities.getOutput().orderItems,
+        totalPaymentDue: c2.getOutput().totalPaymentDue,
+      }),
```
nickevansuk, line 101: It's great how clear this makes everything

packages/openactive-integration-tests/test/features/approval/minimal-proposal/implemented/accept-proposal-book-test.js
```diff
@@ -42,50 +42,351 @@ FeatureHelper.describeFeature(module, {
...
+      createActionFn: () => ({
+        type: 'test:SellerAcceptOrderProposalSimulateAction',
+        objectType: 'OrderProposal',
+        objectId: p.getOutput().orderId,
+      }),
+      requestHelper,
+    });
+    const collectOrderFeedUpdate = OrderFeedUpdateFlowStage.createCollector({
+      getInputState: () => ({
+        getOrderFromOrderFeedPromise: listenForOrderFeedUpdate.getOutput().getOrderFromOrderFeedPromise,
```
nickevansuk, line 125: Also if merging states, and having two collectors created, they would overwrite each other

nickevansuk (commented):
Some thoughts

packages/openactive-integration-tests/test/features/approval/minimal-proposal/implemented/accept-proposal-book-test.js
```diff
@@ -42,50 +42,351 @@ FeatureHelper.describeFeature(module, {
...
+    //
+    // ```
+    // const p = PFlowStage.create({
+    //   ...defaultFlowStageParams,
+    //   getInputState: () => ({
+    //     orderItems: [/** ... */],
+    //     totalPaymentDue: 3.1,
+    //   }),
+    //   prerequisite: c2,
+    // });
```
nickevansuk, line 175: This is great

packages/openactive-integration-tests/test/features/approval/minimal-proposal/implemented/accept-proposal-book-test.js
```diff
@@ -42,50 +42,351 @@ FeatureHelper.describeFeature(module, {
...
+    //
+    // ```
+    // getInputState: FlowStageUtils.getMergedInputState(c2, () => ({ totalPaymentDue: -1.5 }))
+    // ```
+
+    // OrderFeedUpdateFlowStage.wrap simply reduces the order feed update listener/collector boilerplate
+    // (name improvements welcomed :P).
+    // The returned tuple is [wrappedStage, collectOrderFeedUpdateStage].
+    // Yes, the listenForOrderFeedUpdateStage isn't returned as it's not needed for writing tests.
+    const [simulateSellerApproval, orderFeedUpdate] = OrderFeedUpdateFlowStage.wrap({
```
nickevansuk, line 188: This is very clear too, in terms of what's happening where

packages/openactive-integration-tests/test/features/approval/minimal-proposal/implemented/accept-proposal-book-test.js
```diff
@@ -42,50 +42,351 @@ FeatureHelper.describeFeature(module, {
...
+    });
+    const p = PFlowStage.create({
+      ...defaultFlowStageParams,
+      ...FlowStageUtils.prerequisiteAndMergeInputs(c1), // will use totalPaymentDue from c1
+    });
+    const c2 = C2FlowStage.create({
+      ...defaultFlowStageParams,
+      prerequisite: p, // happens after p..
+      getInputState: FlowStageUtils.getMergedInputState(c1), // ..but uses state from c1
+    });
```
nickevansuk, line 269: Also very clear

nickevansuk (commented):
Basically all looks good, just have some uncertainty about merging by default

nickevansuk:
maybe it’s unavoidable if we want to keep the tests simple… but is there any way we can avoid the situation where an invalid response could “inject” a value into the state that gets used in an unexpected way? e.g. if we're merging C1 and C2, and C2 returns null for totalPaymentDue, but we use the value from C1 as a fallback due to merging, the tests will produce a false positive...?

packages/openactive-integration-tests/test/features/approval/minimal-proposal/implemented/accept-proposal-book-test.js
lukehesluke: @nickevansuk I'm not quite sure what you mean. In what way would they overwrite each other?

```js
const listenForOrderFeedUpdate = ...
const someOtherStage = ...
const collectOrderFeedUpdate1 = OrderFeedUpdateFlowStage.createCollector({
  // gets getOrderFromOrderFeedPromise from listenForOrderFeedUpdate
  getInputState: FlowStageUtils.getMergedInputState(someOtherStage),
  ...
})
const collectOrderFeedUpdate2 = OrderFeedUpdateFlowStage.createCollector({
  // gets getOrderFromOrderFeedPromise from listenForOrderFeedUpdate.
  // it would then `await` the promise, as does collectOrderFeedUpdate1, which
  // will then work in the same way
  getInputState: FlowStageUtils.getMergedInputState(collectOrderFeedUpdate1),
  ...
})
```

lukehesluke:
> e.g. if we're merging C1 and C2, and C2 returns null for totalPaymentDue, but we use the value from C1 as a fallback due to merging, the tests will produce a false positive...?

In that particular example, the end result would be a `null` as `null`s should be considered values by the merge.

But I see what you mean. If there was a stage which simply didn't set a value in its state, a previous stage's value will be used instead, which might be unexpected. This is avoided if every stage is sure to set `null` for any value that it should set in its output but doesn't have a value for.

lukehesluke:
Maybe the merging is an unnecessary convenience. I wonder if there are any reasons why merging is preferred as opposed to just writing it out in full each time .. 🤔

nickevansuk:
Mmm I guess it feels a little like some kind of bus implementation, but without the protection to ensure data is sent and received by the right services on the bus...

PR #265, opened 2020-10-22
feature(prepayments)

lukedawilson:
ref impl: https://github.com/openactive/OpenActive.Server.NET/pull/56


.vscode/launch.json
```diff
@@ -5,14 +5,15 @@
...
         {
-            "name": "integration-tests - this test file",
+            "name": "integration-tests - individual",
             "request": "launch",
             "runtimeArgs": [
                 "run-script",
                 "start-tests",
                 "--",
                 "--runInBand",
                 "${relativeFile}",
```
nickevansuk, line 15: This was updated to run the current test that's open in VS Code - do we still need the 'individual' option here?

packages/openactive-integration-tests/config/default.json
```diff
@@ -53,8 +53,8 @@
     "business-to-business-tax-calculation-gross": null,
     "offer-overrides": null,
     "dynamic-payment": null,
-    "prepayment-optional": null,
-    "prepayment-unavailable": null,
+    "prepayment-optional": true,
+    "prepayment-unavailable": true,
```
nickevansuk, line 57: Based on the discussion, we should add an extra feature named `prepayment-required`, which is where the "not implemented" tests sit that use `TestOpportunityBookableFreePrepaymentUnspecified` and `TestOpportunityBookablePaidPrepaymentUnspecified`, and where the implemented tests sit that use `TestOpportunityBookablePaidPrepaymentRequired` and `TestOpportunityBookableFreePrepaymentRequired`

That test should be a prerequisite for both of these to run

Given this is the second time the idea of a prereq feature has come up (leasing was the first), I'll make a note to add prereqs as a feature of the test framework

packages/openactive-integration-tests/test/features/advanced-payment/prepayment-optional/opportunity-paid-test.js
```diff
@@ -0,0 +1,75 @@
...
+
+  describe('B', () => {
+    const b = (new B({
+      state, flow, logger,
+    }))
+      .beforeSetup()
+      .successChecks()
+      .validationTests();
+
+    itShouldReturnPrepayment(b, () => state.bResponse);
```
nickevansuk, line 73: `itShouldReturnPrepayment(b, () => state.bResponse);`

This should be checked at C1 and C2 also.

There should be two test files that use the same testOpportunityCriteria: one where we provide payment, and one where we don't, to check the .NET error logic here

nickevansuk (changes requested):
Good start!

.vscode/launch.json
lukedawilson: No, I'll revert it

packages/openactive-integration-tests/test/templates/b-req.js
```diff
@@ -253,6 +263,7 @@ const bReqTemplates = {
   noBrokerName: createNoBrokerNameBReq,
   incorrectTotalPaymentDuePrice: createIncorrectTotalPaymentDuePriceBReq,
   incorrectOrderDueToMissingPaymentProperty: createIncorrectOrderDueToMissingPaymentProperty,
+  incorrectOrderDueToUnnecessaryPaymentProperty: createIncorrectOrderDueToUnnecessaryPaymentProperty,
   incorrectOrderDueToMissingIdentifierInPaymentProperty: createIncorrectOrderDueToMissingIdentifierInPaymentProperty,
```
nickevansuk, line 267: I notice these are named with the prefix `incorrect`... should they be? For `incorrectTotalPaymentDuePrice` the "incorrect" refers to the price, which is intentionally incorrect. However for the three lines above "incorrect" really depends on context... a missing payment property is correct behaviour in some circumstances (e.g. when prepayment is unavailable)

packages/openactive-integration-tests/test/features/payment/common.js
```diff
@@ -0,0 +1,162 @@
+const chakram = require('chakram');
```
nickevansuk, line 1: Note this whole file needs attention in light of the refactor, and likely most if not all of it should be generalised and live somewhere more generally available to all tests?

packages/openactive-integration-tests/documentation/generator.js
```diff
@@ -46,26 +55,29 @@ global.afterEach = () => {};
...
 // Load metadata from all tests
-var tests = fg.sync(pkg.jest.testMatch, { cwd: rootDirectory }).map(function(file) {
-  console.log('Reading: ' + file);
+const tests = fg.sync(pkg.jest.testMatch, { cwd: rootDirectory }).map(function (file) {
+  console.log(`Reading: ${file}`);
   // TODO: Verify that the data actually conforms to the type.
-  return /** @type {TestModuleExports} */(require(`${rootDirectory}${file}`));
+  const data = /** @type {TestModuleExports} */(require(`${rootDirectory}${file}`));
+  const expectedPath = `test/features/${renderFullTestPath(data)}`;
+  chai.expect(expectedPath).to.equal(file, `Expected ${file} to contain metadata matching its path`);
```
nickevansuk: Added a little something to check the feature metadata identifiers match the test file path when the docs are generated (and fixed some lint issues with the file at the same time)

nickevansuk (commented):
@lukehesluke this needs some test framework work here:

https://github.com/openactive/openactive-test-suite/blob/dc8d1c42da8023e38c4a9a323075979f8690c48d/packages/openactive-integration-tests/test/templates/b-req.js#L23-L29

As the above should take into account the `totalPaymentDue.prepayment` value, so that if it is `https://openactive.io/Unavailable`, then `payment` should not be provided at B.

The tests on this branch will fail randomly until this logic is implemented.

Also note I've focussed purely on the test-level refactor, and not on the detail of the framework or the way the tests themselves are written. It looks like much improvement could be made in light of the recent refactor...

packages/openactive-integration-tests/test/features/payment/non-free-opportunities/feature.json
```diff
@@ -12,5 +12,6 @@
       "name": ".NET Tutorial",
       "href": "https://tutorials.openactive.io/open-booking-sdk/quick-start-guide/storebookingengine/day-5-b-and-delete-order"
     }
-  ]
+  ],
+  "requiresOneOfIfImplemented": ["prepayment-optional", "prepayment-required",  "prepayment-unavailable"]
```
lukehesluke, line 16: @nickevansuk Note: This property is not actually used anywhere in the code. So, for now, it is purely illustrative.

We could add validation of this property in a later PR

packages/openactive-integration-tests/test/features/payment/prepayment-optional/feature.json
```diff
@@ -1,10 +1,11 @@
...
-  "category": "advanced-payment",
+  "category": "payment",
   "identifier": "prepayment-optional",
   "name": "prepayment optional",
   "description": "Support for booking with optional payment",
   "explainer": "",
   "specificationReference": "https://www.openactive.io/open-booking-api/EditorsDraft/#booking-without-payment",
   "required": false,
-  "coverageStatus": "none"
+  "coverageStatus": "none",
```
lukehesluke, line 9: note to self: Why is `coverageStatus: none`? Probably just a mistake

packages/openactive-integration-tests/test/features/payment/prepayment-required/feature.json
```diff
@@ -0,0 +1,11 @@
+{
+  "category": "payment",
+  "identifier": "prepayment-required",
+  "name": "prepayment required",
+  "description": "Support for booking with required payment",
+  "explainer": "",
+  "specificationReference": "https://www.openactive.io/open-booking-api/EditorsDraft/#booking-without-payment",
+  "required": false,
+  "coverageStatus": "none",
```
lukehesluke, line 9: again, why is coverageStatus = none?

packages/openactive-integration-tests/test/features/payment/prepayment-unavailable/feature.json
```diff
@@ -1,10 +1,11 @@
...
+  "category": "payment",
   "identifier": "prepayment-unavailable",
-  "name": "prepayment not required",
+  "name": "prepayment unavailable",
   "description": "Support for booking without payment (reservation only)",
   "explainer": "",
   "specificationReference": "https://www.openactive.io/open-booking-api/EditorsDraft/#booking-without-payment",
   "required": false,
-  "coverageStatus": "none"
+  "coverageStatus": "none",
```
lukehesluke, line 9: hmmm

packages/openactive-integration-tests/test/features/payment/prepayment-unavailable/implemented/opportunity-paid-incomplete-payment-details-test.js
```diff
@@ -0,0 +1,20 @@
...
+const expectedPrepayment = 'https://openactive.io/Unavailable';
+const expectedError = 'UnnecessaryPaymentDetailsError';
+const bReqTemplateRef = 'incorrectOrderDueToMissingIdentifierInPaymentProperty';
+
+FeatureHelper.describeFeature(module, {
+  testCategory: 'payment',
+  testFeature: 'prepayment-unavailable',
+  testFeatureImplemented: true,
+  testIdentifier: 'opportunity-paid-incomplete-payment-details',
+  testName: 'IncompletePaymentDetailsError must be returned in the case that payment details are not supplied',
```
lukehesluke, line 15: this doesn't sound right

packages/openactive-integration-tests/test/features/payment/prepayment-unavailable/implemented/opportunity-paid-incomplete-payment-details-test.js
```diff
@@ -0,0 +1,20 @@
...
+  testDescription: 'An unsuccessful end to end booking, because identifier is missing in the payment property in B request.',
+  testOpportunityCriteria,
+  controlOpportunityCriteria,
+},
+errorTests(expectedPrepayment, expectedError, bReqTemplateRef));
```
lukehesluke, line 20: I don't think this test makes any sense. It attempts, when prepayment is unavailable, to make a payment, but with a missing `identifier`. The payment property itself is erroneous, whether or not it's missing an identifier. Because that case is covered by packages/openactive-integration-tests/test/features/payment/prepayment-unavailable/implemented/opportunity-paid-unnecessary-payment-error-test.js, I think this test can safely be deleted

lukehesluke: I deleted it

packages/openactive-integration-tests/test/templates/b-req.js
lukehesluke: @nickevansuk Totally agreed. I renamed the above one to `noPayment`. It's just missing a payment property. Whether that's incorrect or not depends on the test

packages/openactive-integration-tests/test/features/payment/non-free-opportunities/feature.json
nickevansuk: Good spot, have added: https://github.com/openactive/openactive-test-suite/issues/266

packages/openactive-integration-tests/test/features/payment/prepayment-optional/feature.json
nickevansuk: yes, agreed, likely a mistake

packages/openactive-integration-tests/test/features/payment/prepayment-required/feature.json
nickevansuk: also an error

packages/openactive-integration-tests/test/features/payment/prepayment-unavailable/feature.json
nickevansuk: also error

packages/openactive-integration-tests/test/features/payment/prepayment-unavailable/implemented/opportunity-paid-incomplete-payment-details-test.js
nickevansuk: Ah, so this test is supposed to test to ensure that if the identifier is omitted an `UnnecessaryPaymentDetailsError` is returned, instead of a `IncompletePaymentDetailsError`. The latter could be returned when the identifier is missing, even though the whole payment object is unnecessary, if the booking implementation under test has its error handling logic the wrong way around).

Noting the metadata was erroneously inconsistent here with the test itself.

lukehesluke: Okay, I'll add this test in a separate PR and update the metadata.

Issue #270, opened 2020-10-22
Add helper for C1, C2, B boilerplate in simple case


nickevansuk:
## Helper function for simple case
```javascript
function initialiseSimpleC1C2BFlow(orderItemCriteriaList, logger) {
  const requestHelper = new RequestHelper(logger);

  // ## Initiate Flow Stages
  const defaultFlowStageParams = FlowStageUtils.createDefaultFlowStageParams({ requestHelper, logger });
  const fetchOpportunities = new FetchOpportunitiesFlowStage({
    ...defaultFlowStageParams,
    orderItemCriteriaList,
  });
  const c1 = new C1FlowStage({
    ...defaultFlowStageParams,
    prerequisite: fetchOpportunities,
    getInput: () => ({
      orderItems: fetchOpportunities.getOutput().orderItems,
    }),
  });
  const c2 = new C2FlowStage({
    ...defaultFlowStageParams,
    prerequisite: c1,
    getInput: () => ({
      orderItems: fetchOpportunities.getOutput().orderItems,
    }),
  });
  const b = new BFlowStage({
    ...defaultFlowStageParams,
    prerequisite: c2,
    getInput: () => ({
      orderItems: fetchOpportunities.getOutput().orderItems,
      totalPaymentDue: c2.getOutput().totalPaymentDue,
    }),
  });

  return {c1, c2, b, requestHelper}
}
```

## Usage

```javascript
var {c1, c2, b} = initialiseSimpleC1C2BFlow(orderItemCriteriaList, requestHelper, logger);
```

lukehesluke:
Was just thinking what if it's the same boilerplate c1/c2/b but the test just needs a slight tweak e.g. setting templateRef for c2

in which case, how about this, which makes it easier to drag and drop more complicated flow stages - with the sacrifice of additional boilerplate

# Usage

```js
  const requestHelper = new RequestHelper(logger); // we could also consider injecting this at describeFeature()
  // ## Initiate Flow Stages
  const defaultFlowStageParams = FlowStageUtils.createDefaultFlowStageParams({ requestHelper, logger });
  const fetchOpportunities = FlowStageRecipes.simpleFetchOpportunities(defaultFlowStageParams, orderItemCriteriaList);
  const c1 = FlowStageRecipes.simpleC1({ defaultFlowStageParams, fetchOpportunities });
  const c2 = FlowStageRecipes.simpleC2({ defaultFlowStageParams, fetchOpportunities, c1 });
[...]

PR #285, opened 2020-10-27
chore: Create FlowStageRecipes.initialiseSimpleC1C2BFlow()

lukehesluke:
for simple FetchOpportunities -> C1 -> C2 -> B flows

----

closes https://github.com/openactive/openactive-test-suite/issues/270

This function is tested in: https://github.com/openactive/openactive-test-suite/pull/287

### DIFF
--- a/packages/openactive-integration-tests/test/helpers/flow-stages/flow-stage-recipes.js
+++ b/packages/openactive-integration-tests/test/helpers/flow-stages/flow-stage-recipes.js
@@ -71,7 +71,11 @@ const FlowStageRecipes = {
     });
 
     return {
-      c1, c2, b, requestHelper,
+      requestHelper,
+      fetchOpportunities,
+      c1,
+      c2,
+      b,
     };
   },
 };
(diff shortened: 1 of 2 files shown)


packages/openactive-integration-tests/test/helpers/flow-stages/flow-stage-recipes.js
```diff
@@ -0,0 +1,85 @@
...
+
+const FlowStageRecipes = {
+  /**
+   * Initialise Flow Stages for a simple FetchOpportunities -> C1 -> C2 -> B flow.
+   *
+   * Rather than setting custom input for each stage, the input is just fed automatically
+   * from the output of previous stages.
+   *
+   * DO NOT USE THIS FUNCTION if you want to use custom inputs for each stage (e.g.
+   * to create erroneous requests).
```
nickevansuk, line 32: Perhaps clarify "create erroneous requests" doesn't mean "create requests that return errors", but rather "create malformed requests"?

nickevansuk (approved):
Approved with one minor comment

packages/openactive-integration-tests/test/helpers/flow-stages/flow-stage-recipes.js
lukehesluke: @nickevansuk could it not mean both? e.g. in core/agent-broker/implemented/customer-not-included-test.js, a request is sent to B for which an error response is expected to be returned (therefore "create requests that return errors"). You could also say that the request is malformed as `customer` is supposed to be in the request.

Alternatively, if all the spaces are used up, making a B request is not a malformed request? But it will return an error

PR #286, opened 2020-10-27
feature(cancellation-window)

lukedawilson:
ref impl: https://github.com/openactive/OpenActive.Server.NET/pull/69


packages/test-interface-criteria/src/criteria/criteriaUtils.js
```diff
@@ -84,6 +84,21 @@ function mustBeWithinBookingWindow(offer, opportunity, options) {
...
+  if (!offer || !offer.latestCancellationBeforeStartDate) {
+    return null; // Required for validation step
+  }
+
+  const start = moment(opportunity.startDate);
+  const duration = moment.duration(offer.latestCancellationBeforeStartDate);
+
+  const valid = !start.subtract(duration).isBefore(options.harvestStartTime);
+  return valid;
+}
```
lukehesluke, line 100: @nickevansuk if latestCancellationBeforeStartDate is 2 hours or less, then there will never be an opportunity that satisfies TestOpportunityBookableCancellableWithinWindow, because it uses the `startDateMustBe2HrsInAdvance` constraint (from InternalCriteriaFutureScheduledOpportunity) and this constraint

Therefore, should we suggest, in the Test Interface docs, that they make sure that latestCancellationBeforeStartDate is some number higher than 3 hours? Not sure how we could explain it though.. This is an edge case as I'm sure this value is normally something like 12/24/48 hours. So we could alternatively just leave a code comment for now

lukehesluke (approved):

PR #287, opened 2020-10-27
refactor: Add describeRunAndCheckIsValid() function for FlowStages which are expected to error

lukehesluke:
Also, fix use of template for c1,c2,b, which have not taken into account new brokerRole positional arg in RequestHelper

-----

Depends on https://github.com/openactive/openactive-test-suite/pull/285

In order to test that this worked, I tried it out on an existing test, `customer-not-included`. It works. Hooray.

### DIFF
--- a/packages/openactive-integration-tests/test/helpers/flow-stages/b.js
+++ b/packages/openactive-integration-tests/test/helpers/flow-stages/b.js
@@ -35,7 +35,7 @@ async function runB({ templateRef, uuid, sellerId, orderItems, totalPaymentDue, 
     totalPaymentDue,
     orderProposalVersion,
   };
-  const response = await requestHelper.putOrder(uuid, params, templateRef);
+  const response = await requestHelper.putOrder(uuid, params, null, templateRef);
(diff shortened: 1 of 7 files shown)


packages/openactive-integration-tests/test/helpers/flow-stages/b.js
```diff
@@ -35,7 +35,7 @@ async function runB({ templateRef, uuid, sellerId, orderItems, totalPaymentDue,
     totalPaymentDue,
     orderProposalVersion,
   };
-  const response = await requestHelper.putOrder(uuid, params, templateRef);
+  const response = await requestHelper.putOrder(uuid, params, null, templateRef);
```
lukehesluke, line 38: There's a new `brokerRole` arg in these 3 requestHelper functions, so templates weren't working 😧

nickevansuk: Good catch!

nickevansuk (approved):

PR #304, opened 2020-11-02
feat(prepayment-unavailable): Add opportunity-paid-incomplete-payment-details test

lukehesluke:


packages/openactive-integration-tests/test/templates/b-req.js
```diff
@@ -44,7 +44,21 @@ const { createPaymentPart, isPaidOpportunity, isPaymentAvailable } = require('./
...
+ * B request template don't match up.
+ *
+ * @param {BReqTemplateRef} templateName
+ * @param {BReqTemplateData} data
+ */
+function assertPaymentIsUnvailable(templateName, data) {
+  if (isPaymentAvailable(data)) {
+    throw new Error(`${templateName} B request incorrectly used for an Order for which prepayment is available. Consider using another B request template or a different OpportunityCriteria`);
   }
 }
```
lukehesluke, line 63: I introduced these assertions in the previous PR because I was worried that, with the scale of templates now available, it would be easy to use one in a test without realising that it doesn't work for that kind of test (and therefore gives false positives/negatives). Therefore, these assertions add a small amount of sanity checking

Perhaps all the templates shouldn't live in one centralized place. But that's a problem for later

PR #310, opened 2020-11-04
Feature/order deletion tests

### LINKED ISSUE #309: order-deletion - multiple - order deletion tests
**Feature**

order-deletion

**Use Case**

- (orders-updated-then-deleted) Run C1, C2 and B, then run U, then run Order Deletion, and check Orders feed that Order has been deleted
- (order-delete-idempotent) {new} Order Delete is idempotent - run C1, C2, and B then Order Delete twice - the Order Delete must return 204 in both cases
- (order-quote-delete-idempotent) {new} Order Quote Delete is idempotent - run C1, C2 then Order Quote Delete twice - the Order Quote Delete must return 204 in both cases

[Open Booking API spec](https://www.openactive.io/open-booking-api/EditorsDraft/#order-deletion).

NemanjaStefanovic:
_current state (12th Nov)_

- REQUIRED .NET WORK: https://github.com/openactive/OpenActive.Server.NET/issues/87
  - For now, the relevant part of the test has been `describe.skip`d
- GoneError: Let's figure this out separately: https://github.com/openactive/openactive-test-suite/issues/323

-----

_The old state as of 3rd Nov:_

This PR contains 2 completed tests and 1 partially implemented from row 9 order-deletion:
Completed:
4. (order-delete-idempotent) {new} Order Delete is idempotent - run C1, C2, and B then Order Delete twice - the Order Delete must return 204 in both cases
5. (order-quote-delete-idempotent) {new} Order Quote Delete is idempotent - run C1, C2 then Order Quote Delete twice - the 
Order Quote Delete must return 204 in both cases

Partially completed:
1. (orders-updated-then-deleted) Run C1, C2 and B, then run U, then run Order Deletion, and check Orders feed that Order has been deleted.
I didn't know how to implement this last check, I used @lukehesluke's approach where I made Flow stage from order deletion and used his approach of wrapping action to detect changes in Orders feed, but change is not registered as order is deleted not changed. I also tried calling GetOrderStatus endpoint, but it is only for active Orders (if it is deleted in db it won't show).
So I left this approach of making new flow stage and wrapping call to detect changes to order feed just in case.

Other two:
2. Test for GoneError - is missing because I didn't know how to trigger soft delete instead of hard delete (I think by default booking system now does the hard delete. @nickevansuk If you know how this is done you can tag me in comment and I will implement it.).
3. (unknown-order) {new} Test for UnknownOrderError at Order Deletion endpoint - already exists.

Closes #309


packages/openactive-broker-microservice/app.js
```diff
@@ -813,7 +813,7 @@ function processOpportunityItem(item) {
 /** @type {RpdePageProcessor} */
 function monitorOrdersPage(rpde) {
   rpde.items.forEach((item) => {
-    if (item.data && item.id && orderResponses[item.id]) {
+    if (item.id && orderResponses[item.id]) {
```
lukehesluke, line 816: Required to allow deletes to be picked up by the `/get-order/:orderUuid` endpoint

lukehesluke (approved):

PR #343, opened 2021-03-22
feat: Better dataset site errors

nickevansuk:


packages/openactive-broker-microservice/app.js
```diff
@@ -884,17 +891,29 @@ async function startPolling() {
 
   const harvesters = [];
 
+  const isParentFeed = {
```
civsiv, line 894: Totally optional as it doesn't affect functionality, just readability:

Calling this `isParentFeed` feels slightly misleading as it's being used as a 3-branch conditional below, and then just as a list of feeds in error statement.

Maybe slightly more readable would be:
```
const additionalTypeParentNonParentMap = {
https://openactive.io/SessionSeries': PARENT_ADDITIONAL_FEED,
....
```
and 
```
if (additionalTypeParentNonParentMap[dataDownload.additionalType] === PARENT_ADDITIONAL_FEED)
```
etc

imo `Object.keys(isParentFeed).map((x)` is a bit misleading as we're using it for all feeds, not just parent ones

civsiv (approved):

packages/openactive-broker-microservice/app.js
nickevansuk: Oh sure - is `PARENT_ADDITIONAL_FEED` a `Symbol` here?

nickevansuk: (Merging now for speed)

civsiv: >Oh sure - is PARENT_ADDITIONAL_FEED a Symbol here?

Yeh exactly

PR #344, opened 2021-03-23
feat: Validation of opportunity feeds

### LINKED ISSUE #89: Run RPDE validation rules
Run RPDE validation rule across each page, to check they're consistent

nickevansuk:
Add inline validation to feeds, which closes #89

### DIFF
--- a/packages/openactive-broker-microservice/app.js
+++ b/packages/openactive-broker-microservice/app.js
@@ -90,7 +90,7 @@ async function validateItem(body) {
    */
   const optionsWithRemoteJson = {
     loadRemoteJson: true,
-    remoteJsonCachePath: '../openactive-integration-tests/tmp',
+    remoteJsonCachePath: VALIDATOR_TMP_DIR,
     remoteJsonCacheTimeToLive: 3600,
     validationMode: 'BookableRPDEFeed',
   };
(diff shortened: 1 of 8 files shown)


packages/openactive-broker-microservice/app.js
```diff
@@ -62,6 +71,108 @@ const startTime = new Date();
...
+const validationResults = new Map();
+
+/**
+ * Use OpenActive validator to validate the opportunity
+ *
+ * @param {any} data opportunity JSON-LD object
+ */
+async function storeValidationResults(data) {
+  const id = data['@id'] || data.id;
+  const errors = await validateItem(data, false);
```
civsiv, line 111: If `isOrdersFeed` is always false, is it needed? Or are you anticipating validating Orders Feeds too? If so, maybe a comment to explain?

packages/openactive-broker-microservice/app.js
```diff
@@ -62,6 +71,108 @@ const startTime = new Date();
...
+}
+
+const validationResults = new Map();
+
+/**
+ * Use OpenActive validator to validate the opportunity
+ *
+ * @param {any} data opportunity JSON-LD object
+ */
+async function storeValidationResults(data) {
```
civsiv, line 109: This function both validates and stores results, so maybe should be called something like `validateAndStoreValidationResults`?

civsiv (approved):

packages/openactive-broker-microservice/app.js
nickevansuk: Good point, YAGNI, have removed

packages/openactive-broker-microservice/app.js
nickevansuk: Good shout, have updated

PR #365, opened 2021-04-22
feat: change requests to be IdReference

civsiv:
Implements https://github.com/openactive/open-booking-api/issues/191


lukehesluke (approved):
I don't really know the context but lgtm. I'm surprised that this hasn't caused any CI issues. I was assuming that some of the tests would rely on `orderedItem["@type"]` but apparently not 🙌

nickevansuk (approved):

PR #386, opened 2021-04-27
feat: Delete check for proposal

nickevansuk:
Add test to ensure that OrderProposal is always deleted after B, as per the spec.

This is also important to ensure that this change in the Orders Feed is collected explicitly before the next call is made, to avoid a race condition where the "delete" is actually picked up by any subsequent listeners.

### DIFF
--- a/packages/openactive-integration-tests/test/features/payment/common.js
+++ b/packages/openactive-integration-tests/test/features/payment/common.js
@@ -73,6 +73,9 @@ function successTests(expectedPrepayment, bookReqTemplateRef) {
     FlowStageUtils.describeRunAndCheckIsSuccessfulAndValid(bookRecipe, () => {
       if (bookRecipe.p) {
         itShouldHavePrepayment(expectedPrepayment, () => bookRecipe.p.getOutput().httpResponse.body);
+        it('should have state: deleted', () => {
+          expect(bookRecipe.orderFeedUpdateAfterDelete.getOutput().httpResponse.body.state).to.equal('deleted');
+        });
       }
       itShouldHavePrepayment(expectedPrepayment, () => bookRecipe.b.getOutput().httpResponse.body);
     });
--- a/packages/openactive-integration-tests/test/features/payment/common.js
+++ b/packages/openactive-integration-tests/test/features/payment/common.js
@@ -73,9 +73,6 @@ function successTests(expectedPrepayment, bookReqTemplateRef) {
     FlowStageUtils.describeRunAndCheckIsSuccessfulAndValid(bookRecipe, () => {
       if (bookRecipe.p) {
         itShouldHavePrepayment(expectedPrepayment, () => bookRecipe.p.getOutput().httpResponse.body);
-        it('should have state: deleted', () => {
-          expect(bookRecipe.orderFeedUpdateAfterDelete.getOutput().httpResponse.body.state).to.equal('deleted');
-        });
       }
       itShouldHavePrepayment(expectedPrepayment, () => bookRecipe.b.getOutput().httpResponse.body);
     });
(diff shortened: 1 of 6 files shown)


packages/openactive-integration-tests/test/helpers/flow-stages/flow-stage-recipes.js
```diff
@@ -204,28 +204,38 @@ const FlowStageRecipes = {
...
         },
       });
+
       return new BookRecipe({
         firstStage: p,
         p,
         simulateSellerApproval,
         orderFeedUpdateCollector,
         b,
+        orderFeedUpdateAfterDelete,
```
lukehesluke, line 238: @nickevansuk I'm confused by what's happening here. Why is there an Order Feed update happening after B?

And why is it called `orderFeedUpdateAfterDelete`? It doesn't look like anything's being deleted?

packages/openactive-integration-tests/test/features/payment/common.js
```diff
@@ -73,6 +73,9 @@ function successTests(expectedPrepayment, bookReqTemplateRef) {
     FlowStageUtils.describeRunAndCheckIsSuccessfulAndValid(bookRecipe, () => {
       if (bookRecipe.p) {
         itShouldHavePrepayment(expectedPrepayment, () => bookRecipe.p.getOutput().httpResponse.body);
+        it('should have state: deleted', () => {
+          expect(bookRecipe.orderFeedUpdateAfterDelete.getOutput().httpResponse.body.state).to.equal('deleted');
+        });
```
lukehesluke, line 78: And then why is it the successful payment test which is checking for this deleted item?

lukehesluke:
You need to add it to the `FlowStageUtils.describeRunAndRunChecks(..)`  function in order to get that Flow stage to run but I'm confused as to why it's there in the 1st place

packages/openactive-integration-tests/test/features/payment/common.js
lukehesluke: Perhaps it should be in one of the `minimal-proposal` tests?

lukehesluke: You could do this by splitting out `FlowStageRecipes.book` into ``FlowStageRecipes.bookSimple` and `FlowStageRecipes.bookApproval`. `book` just returns one or the other depending on the order item criteria given to it.

This way, you can have the minimal-proposal tests use `FlowStageRecipes.bookApproval`, which means that they get the benefit of using the book recipe, which includes things like this new Order Feed update listener while still being explicit about only using the approval flow

packages/openactive-integration-tests/test/helpers/flow-stages/flow-stage-recipes.js
```diff
@@ -204,28 +204,38 @@ const FlowStageRecipes = {
...
+              totalPaymentDue: p.getOutput().totalPaymentDue,
+              orderProposalVersion: p.getOutput().orderProposalVersion,
+              prepayment: p.getOutput().prepayment,
+            };
+          },
+        })),
+        orderFeedUpdateParams: {
+          ...defaultFlowStageParams,
+          prerequisite: orderFeedUpdateCollector,
+          testName: 'Orders Feed (after B)',
```
lukehesluke, line 228:
```suggestion
          testName: 'OrderProposal Deletion in Feed (after B)',
```

packages/openactive-integration-tests/test/helpers/flow-stages/book-recipe.js
```diff
@@ -18,13 +18,15 @@ class BookRecipe {
...
+   * @param {OrderFeedUpdateCollectorType | null} [args.orderFeedUpdateAfterDelete]
    */
-  constructor({ b, p, simulateSellerApproval, orderFeedUpdateCollector, firstStage }) {
+  constructor({ b, p, simulateSellerApproval, orderFeedUpdateCollector, firstStage, orderFeedUpdateAfterDelete }) {
     this.b = b;
     this.p = p;
     this.simulateSellerApproval = simulateSellerApproval;
     this.orderFeedUpdateCollector = orderFeedUpdateCollector;
     this.firstStage = firstStage;
+    this.orderFeedUpdateAfterDelete = orderFeedUpdateAfterDelete;
```
lukehesluke, line 29: Maybe `orderFeedUpdateAfterDeleteProposal`

packages/openactive-integration-tests/test/features/payment/common.js
nickevansuk: ah maybe I wasn't following our chat earlier... the test for the delete was to check that the proposal got deleted... which should be checked every time?

PR #407, opened 2021-05-06
feat: Create *-without-checkpoints tests

lukehesluke:


packages/openactive-integration-tests/test/features/payment/non-free-opportunities/implemented/opportunity-paid-without-checkpoints-test.js
```diff
@@ -0,0 +1,38 @@
+const { FeatureHelper } = require('../../../../helpers/feature-helper');
```
lukehesluke, line 1: This test unfortunately won't be run by CI at present. We could try and make it testable in CI by having a mode for the Ref Impl that turns off all tax calculations and then running just this (and some other tax-agnostic tests) against that Ref Impl.

Maybe an issue for later?

nickevansuk: Yes good idea having something in .NET that does this

PR #408, opened 2021-05-06
fix: Use failEarlyIf to stop test timeouts waiting for Order Feed updates that are never going to come after failed P

lukehesluke:
## QA

Without the fix, if I force a fail at P, the tests stop and timeout as they wait for that OrderProposal feed update that will never come. This leads to very slow feedback for anyone implementing approvals:

Screenshot 2021-05-06 at 14 37 23

With the fix, later tests fail early:

Screenshot 2021-05-06 at 14 29 19

Screenshot 2021-05-06 at 14 29 56

Much quicker ☺️

### DIFF
--- a/packages/openactive-integration-tests/test/helpers/flow-stages/order-feed-update.js
+++ b/packages/openactive-integration-tests/test/helpers/flow-stages/order-feed-update.js
@@ -46,9 +46,14 @@ const { FlowStageUtils } = require('./flow-stage-utils');
  * @param {string} args.uuid
  * @param {RequestHelperType} args.requestHelper
  * @param {OrderFeedType} args.orderFeedType
+ * @param {() => boolean} args.failEarlyIf
  * @returns {Promise<ListenerOutput>}
  */
-async function runOrderFeedListener({ uuid, requestHelper, orderFeedType }) {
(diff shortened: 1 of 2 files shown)


packages/openactive-integration-tests/test/helpers/flow-stages/order-feed-update.js
```diff
@@ -203,6 +230,12 @@ const OrderFeedUpdateFlowStageUtils = {
       uuid: orderFeedUpdateParams.uuid,
       logger: orderFeedUpdateParams.logger,
       orderFeedType,
+      failEarlyIf: () => {
+        if (failEarlyIf()) { return true; }
+        // Don't bother collecting if the wrapped stage fails.
+        if ((wrappedStage.getOutput()?.httpResponse?.response?.statusCode ?? 200) >= 400) { return true; }
```
lukehesluke, line 236: Also added this bit to automatically fail if the wrapped stage fails. This needs to be done automatically rather than with the `failEarlyIf` arg, because the wrapped stage isn't accessible to the caller of `OrderFeedUpdateFlowStageUtils.wrap(..)` because `OrderFeedUpdateFlowStageUtils.wrap(..)` creates the wrapped stage.

This also means that this doesn't just fix timeouts after P fails. It should fix all fail -> listen for Order Feed update timeouts

PR #426, opened 2021-05-24
add Events to config

civsiv:


nickevansuk:
Useful for reference, but not currently implemented

Issue #441, opened 2021-06-25
Ensure Booking Partner and Seller are securely partitioned


nickevansuk:
There are currently no tests to ensure that correct security is implemented for the Orders feed.

## Additional tests

`booking-partner-authentication` feature should be split into:
- `booking-partner-openid-authentication-for-multiple-sellers` - the `authorization-code-flow` related tests (required if `multiple-sellers` feature is enabled)
- `booking-partner-openid-authentication` - the `client-credentials-flow` tests (required if `openid-authentication-for-multiple-sellers` feature is enabled)
- `booking-partner-authentication` - this is a set of authentication-agnostic tests to check the booking partner partitioning is implemented correctly (always required)
   - `booking-partner-partitioning-for-orders` test
      - Orders from two different bookings partners must not be visible to each other, and UUID must be unique within each booking partner
      - Using **primary** booking partner auth, make a booking using a UUID, cancel it, and check for it in the Orders feed. Using **secondary** booking auth, assert that an Order with the same UUID is not present in the orders feed. Then using **secondary** booking partner auth, make a different booking using the same UUID, cancel it, and check for it in the Orders feed. Both booking should succeed.
   - `booking-partner-partitioning-for-order-proposals` test
      - Order Proposals from two different bookings partners must not be visible to each other, and UUID must be unique within each booking partner
       - (if `minimal-proposal` is implemented) Using **primary** booking partner auth, make a P call using a UUID, approve it, and check for it in the Order Proposals feed. Using **secondary** booking auth, assert that an Order Proposal with the same UUID is not present in the Order Proposals feed. Then using **secondary** booking partner auth, make a different P call using the same UUID, approve it, and check for it in the Orders feed.   

`multiple-sellers` feature should also include:
   - `seller-access-restricted-by-auth` test
      - Credentials for Seller (a) must not provide access make bookings for Seller (b)
      - Using **primary** seller auth, make a call to C1, C2, and P/B for the **secondary** seller, expecting all calls to fail with `InvalidAuthorizationDetailsError` (this is supported on the `coverage/additional-auth-tests` branch)

## Additional test suite features to support these 

To implement the above, recommend additional features in the broker microservice:
1) Allow Orders feed listening to take into account the booking-partner creds to use (primary or secondary, already configured) - this likely means harvesting double the number of Orders feeds (harvesting running for each set of creds, `getOrdersFeedHeader` looks like it could easily be parameterised to support `primary` and `secondary`). 
[...]

PR #449, opened 2021-07-01
feat(broker): /is-order-uuid-present API

lukehesluke:
Which returns whether or not a given Order UUID exists in the specified feed

This is for the `booking-partner-authentication` tests from https://github.com/openactive/openactive-test-suite/issues/441

## QA

I've tested the following scenarios:

1. Turn on broker, open `http://localhost:3000/is-order-uuid-present/orders/primary/abc` before feeds have started harvesting

    This requests waits until broker finishes harvesting feeds and then returns `false` ✅ 
2. Turn on broker, wait for feeds to finish harvesting, open `http://localhost:3000/is-order-uuid-present/orders/primary/abc`

    Immediately returns `false` ✅ 
3. Turn on broker, run the `seller-requested-cancellation` test, which should place an Order in the feed, check the UUID of the generated Order using the test results. The ID was `da32a390-267b-42fd-9748-27600b29837e`. Then, open `http://localhost:3000/is-order-uuid-present/orders/primary/da32a390-267b-42fd-9748-27600b29837e`

    Immediately returns `true` ✅ 
4. After # 3, restart Broker and open `http://localhost:3000/is-order-uuid-present/orders/primary/da32a390-267b-42fd-9748-27600b29837e` before it finishes harvesting feeds.

    Request waits until Broker has finished harvesting feeds and then returns `true` ✅ 
5. After # 3, try requests for different UUIDs and for the `secondary` booking partner.

    Confirm that these all return `false`. ✅ 
6. After # 3, also check that it works for the `order-proposals` feed with one of the approval runs of the same test.

    Again, worked as expected - returned `true` ✅


packages/openactive-broker-microservice/app.js
```diff
@@ -190,18 +193,37 @@ function withOpportunityRpdeHeaders(getHeadersFn) {
 
 /**
  * @param {string} baseUrl
- * @param {string} feedIdentifier
+ * @param {string} feedContextIdentifier
```
lukehesluke, line 196: Trying to improve consistency. So that `feedContextIdentifier` always refers to the string which can be `OrdersFeed (auth:secondary)`. `feedIdentifier` is used separately to refer to just the `OrdersFeed` part

PR #472, opened 2021-12-14
Pull request to store a variety of hacks

nickevansuk:
This PR is temporary, to document some hacks that had previously been applied to test suite in lieu of full feature development


nickevansuk:
These have since been incorporated into the test suite

Issue #473, opened 2021-12-15
Test suite not running on Linux


nickevansuk:
Threading issues are apparently preventing the test suite from running on Linux (Ubuntu 20.04).

Solution is to use a better threading library.

lukehesluke:
I have also seen this issue on a Linux machine

# What happens

1. Run broker
2. Broker Harvests all feeds and then logs that there are validation errors:
3. Open the validation errors page and see this issue:

```
Validation Library Error
Timeout: Did not receive an init message from worker after 10000ms. Make sure the worker calls expose().
```

# The issue

As Nick pointed out, this happens as a result of the `validator-worker.js` validator worker threads (which are spawned by `AsyncValidatorWorker`) not being able to initialise in Linux. Not sure exactly why though

nickevansuk:
I've also seen this on MacOS under heavy load, so it seems to be some kind of race condition? Perhaps the application running locally under test uses resources intensely it could cause the issue on any platform? (in my report above there were several docker containers running on the machine that produced the error)

PR #479, opened 2022-02-17
fix(broker): Ditch threads.js for validation multithreading

lukehesluke:
Fixes https://github.com/openactive/openactive-test-suite/issues/473

What this PR does:

* Removes the threads.js dependency and just uses native worker_threads
* Number of parallel workers is fixed for a given computer - equal to number of cores. Before, this was set to the number of feeds. This should improve performance on a computer which has more cores than feeds and presumably also when a computer has fewer cores (as previously this computer would have been hosting multiple workers per core)
* Rather than enqueueing all items to be validated, items are saved to files (in chunks), which the worker pool then processes.
  * Before, the worst case memory usage might have been holding the entirety of all feeds in memory if validator was slow for some reason (e.g. because the OA context takes 30s to load). This is because it would have been possible to enqueue the entire feed.
  * So, now, items are saved to files and then read one at a time (well, actually: "\<NUMBER OF CORES\> at a time")

## Notes to Reviewer

I recommend starting with:

1. validator-worker-pool.js
2. validator-worker.js
3. validator-inputs.js
4. app.js

## QA

I asked someone who experienced https://github.com/openactive/openactive-test-suite/issues/473 to run this branch, to see if this fixes the issue.

It fixed it!

Since cleaning this up, I have continued to run Broker (`NODE_ENV=dev npm run start-broker`) and see that it works. Here's the results of my latest run:

Screenshot 2022-02-23 at 17 46 10

Screenshot 2022-02-23 at 17 48 10


packages/openactive-broker-microservice/app.js
```diff
@@ -90,54 +91,14 @@ if (REQUEST_LOGGING_ENABLED) {
...
-      state.validationResults.set(errorKey, currentValidationResults);
-    }
-
-    // Keep track of examples of each error, with a preference for newer ones (later in the feed)
-    currentValidationResults.occurrences += 1;
-    currentValidationResults.examples.unshift(id);
-    if (currentValidationResults.examples.length > 5) {
-      currentValidationResults.examples.pop();
-    }
-  }
```
lukehesluke, line 132: This logic is moved into ValidatorWorkerPool._processValidationError pretty much verbatim.

With the one exception that the filtering logic `if (data['@type'] === 'SessionSeries' && // ...` is moved into validator-worker's error filtering logic (`isIgnorableResult(..)`)

packages/openactive-broker-microservice/src/validator/validator-inputs.js
```diff
@@ -0,0 +1,113 @@
+/**
+ * Items are sent to the Validator Worker Pool (VWP) by saving them to files, which the VWP then reads and processes.
+ *
+ * The reason that they are sent as files, rather than, say maintained in memory, is so that there is a hard limit on
+ * memory usage from queuing items to Validator.
+ *
+ * The potential worst case scenario is one in which Validator takes a long time to start the first time e.g. because
```
civsiv, line 7: the worst case of the previous implementation right?

packages/openactive-broker-microservice/src/validator/validator-inputs.js
```diff
@@ -0,0 +1,113 @@
...
+/**
+ * Call this after Validator Worker Pool has stopped.
+ */
+async function cleanUpValidatorInputs() {
+  // Once validator has stopped, we may as well clean this directory to free up storage space
+  await clearValidatorInputTmpDir();
+}
+
+/**
+ * For a given page of items, split out those which are validatable (i.e. ignore deleted ones) into files and save
```
civsiv, line 58: Are deleted items not validated? I vaguely remember seeing deleted items with a `data` object which is obviously wrong and should be flagged? Not even sure the validator validates against that usecase tbh

civsiv (approved):

packages/openactive-broker-microservice/src/validator/validator-inputs.js
```diff
@@ -0,0 +1,113 @@
...
+      item: item.data,
+      validationMode: ITEM_VALIDATION_MODE,
+      feedContextIdentifier,
+    })),
+    /* Items are split into chunks so e.g. a page with 500 items will, with chunk length 100, be split into 5 files
+    with 100 items each.
+
+    Compared to saving the entire page as one input file to Validator Worker Pool, this ensures that:
+
+    - There's a hard limit on memory usage from queueing items for validation, which is ITEM_SIZE * CHUNK_LENGTH * NUM_CORES.
```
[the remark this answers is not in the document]
lukehesluke: No I think this is a very good point. I'll change the wording

packages/openactive-broker-microservice/src/validator/validator-inputs.js
lukehesluke: Yes. Deleted items are not validated. This is the same as the previous logic. I agree it would be better to validate more wholistically but maybe one for another PR? Only reason I can think that it wasn't done before is maybe performance but probably just an oversight

civsiv: Ok sure 👍🏽

packages/openactive-broker-microservice/app.js
```diff
@@ -1357,7 +1274,9 @@ async function processOpportunityItem(item) {
  * @returns {RpdePageProcessor}
  */
 function monitorOrdersPage(orderFeedType, bookingPartnerIdentifier) {
-  return (rpdePage) => {
+  /* Note that the Orders RpdePageProcessor does NOT use validateItemsFn i.e. Orders feed items are not validated
+  LW: I don't know why that is */
```
nickevansuk, line 1278: Great point! As Orders are only created in response to bookings, and the Orders feed is specific to the Broker (i.e. the test suite), and as the test suite validates the Order as part of the test, the idea was that there shouldn't be a need to validate what's already in the feed (as they'd have been validated as part of earlier tests).

In controlled mode, where the test data is wiped before each run, the Orders feed is expected to be empty.

So I guess the question becomes should we validate any Orders from previous bookings in Random mode... perhaps it's worth doing this for completeness?


However it could be that Orders ah

packages/openactive-broker-microservice/app.js
```diff
@@ -534,10 +456,13 @@ async function setFeedIsUpToDate(feedIdentifier) {
...
-        // Stop the validator threads as soon as we've finished harvesting - so only a subset of the results will be validated
-        // Note in some circumstances threads will complete their work before terminating
-        await Promise.all(state.validatorThreadArray.map(async (validator) => validator.terminate));
-        // TODO surely this doesn't work ^ needs to actually call terminate()
+        /* Signal for the Validator Worker Pool that we may stop once the validator timeout has run.
+        Validator is an expensive process and is not completely necessary for Booking API testing. So we put a hard
+        limit on how long it runs for (once all items are harvested).
+        This means that, in some cases, only a subset of the results will be validated.
+        Note that the worker pool will finish its current iteration if it has already reached the timeout. */
+        await validatorWorkerPool.stopWhenTimedOut();
```
nickevansuk, line 464: This is great! Just noting that when in `VALIDATE_ONLY` mode it should instead wait for validation to complete across all items (rather than killing validation early when harvesting is complete)

This might be one for a separate task, but thought I'd flag here in case it was useful to note how this might be achieved while you're in the context?

nickevansuk (approved):

packages/openactive-broker-microservice/app.js
nickevansuk: Could be added as a separate issue if so?

packages/openactive-broker-microservice/app.js
lukehesluke: Thanks @nickevansuk! Silly me, hadn't realised that.

Have fixed this now. If VALIDATE_ONLY=true, then WAIT_FOR_HARVEST=false, in which case validatorTimeout=null, in which case, **ValidatorWorkerPool will not wait for a timeout**

Most of that logic was already in place, but the bold bit wasn't.

Looking at the code though, it seems like this might have been broken before? As, if `validatorTimeoutMs=null`, then in that previous logic, the condition at [async-validator-worker.js:19](https://github.com/openactive/openactive-test-suite/pull/479/files#diff-017ca6e804ec1ceb0b34fc727657feab462cc9055c089ccc172fe2dfb80ca2e1L19), `remainingValidationTimeoutMs > 0` will always be true because `<number> - null` = `<number>` in JS. So assuming that my understanding of what's supposed to happen here is correct, this looks to be fixed!

packages/openactive-broker-microservice/app.js
lukehesluke: Have added as a separate issue (https://github.com/openactive/openactive-test-suite/issues/480) and updated the comment here to explain why it is as it is here

PR #485, opened 2022-05-23
fix: Update puppeteer to avoid test runner hanging

sbscully:
With recent versions of Chrome installed puppeteer would hang in packages/openactive-openid-test-client/lib/browser-automation-for-auth.js.

This required also installing a newer version of TypeScript to avoid type errors.


sbscully:
Closing to restart and try a different approach

PR #489, opened 2022-05-27
feat: B/P detailed errors

nickevansuk:
Changes the tests to expect detailed errors for stage B (and P in approval flow).

There are still quite a few tests failing. I've tried to make sure the only tests that are failing, are failing because the reference implementation is broken (ie the tests should be failing).

Related issue: https://github.com/openactive/open-booking-api/issues/218

(Branch renamed from #488 to pass CI)


nickevansuk:
So @sbscully  the remaining errors are actually to do with the validator: which doesn’t currently differentiate between responses with OrderItem errors, and responses without.

I’ve updated the validator to do this (https://github.com/openactive/data-models/pull/91) and just checking through the PR now.

The main thing this does is add additional validation modes `C1ResponseOrderItemError` , `C2ResponseOrderItemError`, `BResponseOrderItemError` and `PResponseOrderItemError`.

Could we update this PR so that the test suite uses these too? Once done all tests should pass

PR #493, opened 2022-06-07
feat: Add endpoint to generate sample requests

sbscully:
Adds an endpoint /sample-requests that generates valid requests for each endpoint in the simple and approval booking flows. This is used in the [Postman workspace](https://documenter.getpostman.com/view/21015180/Uz5DqdCf) for the developer documentation.


packages/openactive-broker-microservice/src/sample-requests.js
```diff
@@ -0,0 +1,237 @@
...
+      "valueExpr": {
+        "@type": "test:OptionNodeConstraint",
+        "datatype": "oa:RequiredStatusType",
+        "blocklist": [
+          "https://openactive.io/Unavailable"
+        ]
+      }
+    }
+  ]
+};
```
nickevansuk, line 87: This shouldn't be required, as `test:testOpportunityCriteria` and `test:testOpenBookingFlow` should be sufficient

packages/openactive-broker-microservice/src/sample-requests.js
```diff
@@ -0,0 +1,237 @@
...
+        "orderItemStatus": "https://openactive.io/CustomerCancelled"
+      }
+    ]
+  }
+}
+
+function buildSampleRequests(opportunity) {
+  const parent = opportunity.superEvent || opportunity.facilityUse;
+  const seller = parent.organizer || parent.provider;
+  const offer = parent.offers[0]
```
nickevansuk, line 221: This should use: https://github.com/openactive/openactive-test-suite/blob/ad0294105bd1dbcaf407fdf87c1380d69ae70c57/packages/openactive-integration-tests/test/helpers/flow-stages/fetch-opportunities.js#L46-L55 rather than just the first `Offer`, as the first `Offer` may not conform with the criteria.

Copying this hint from previous discussions:
> This little bit of logic could either be (a) centralised in the test-interface-criteria package alongside the getRelevantOffers function that it calls, or (b) just duplicated into the broker microservice - for the purposes of the endpoint we’re talking about creating
The opportunity that’s passed in is the full opportunity that you get from the second of the calls we looked at in the test suite output file

nickevansuk (changes requested):
A couple of notes aside from the linting issues.

packages/openactive-broker-microservice/src/sample-requests.js
```diff
@@ -0,0 +1,157 @@
...
+    harvestStartTime: HARVEST_START_TIME,
+  });
+  if (relevantOffers.length === 0) { return null; }
+
+  return relevantOffers[Math.floor(Math.random() * relevantOffers.length)];
+}
+
+function buildSampleRequests(opportunity, criteriaName) {
+  const parent = opportunity.superEvent || opportunity.facilityUse;
+  const seller = parent.organizer || parent.provider;
```
nickevansuk, line 141: For the request this should be the `@id` rather than the whole seller object, so can use this too:

https://github.com/openactive/openactive-test-suite/blob/6f477cfc64f318ba198d93c5f83db31daa38d1cf/packages/openactive-broker-microservice/app.js#L943-L950
```suggestion
  const sellerId = detectSellerId(opportunity);
```

packages/openactive-broker-microservice/app.js
```diff
@@ -1025,6 +1027,28 @@ app.post('/assert-unmatched-criteria', function (req, res) {
   }
 });
 
+// put our endpoint here
+app.get('/sample-requests', function (req, res) {
+  // Get random opportunity ID
+  const opportunity = req.body;
+  const opportunityType = detectOpportunityType(opportunity);
+  const sellerId = opportunity.superEvent?.organizer || opportunity.facilityUse?.provider;
```
nickevansuk, line 1035: This should make use of: https://github.com/openactive/openactive-test-suite/blob/6f477cfc64f318ba198d93c5f83db31daa38d1cf/packages/openactive-broker-microservice/app.js#L943-L950



```suggestion
  const sellerId = detectSellerId(opportunity);
```

nickevansuk (changes requested):
Further notes - apologies for missing these in the previous review

nickevansuk (approved):
Assuming it works as expected, all looks good to me!

PR #500, opened 2022-06-21
fix: Update `endpointUrl` case to match latest

nickevansuk:
Update in line with https://github.com/openactive/data-models/pull/99


nickevansuk:
Closed as duplicate of #504

PR #513, opened 2022-12-30
force email as lowercase

eliasfernandez:
As the email is a composition of the first name (first letter in upper case), it will fail if the system tries to store it in lowercase


nickevansuk:
Assume that test suite fails as the email address returned doesn't match the one that's sent in the request?

In which case should this PR actually make that check case-insensitive, rather than changing the test data it's using?

Reason being that the test data is valid, it's the check that's overly prescriptive based on the RFC

nathansalter (approved):

nickevansuk (approved):

nickevansuk:
Looks good, great work both!

PR #517, opened 2023-01-24
Feature/client credentials for seller

civsiv:


config/.example.single-seller-client-credentials.json
```diff
@@ -0,0 +1,47 @@
...
+    "implementedFeatures": {
+      "multiple-sellers": false,
+      "booking-partner-authentication": false,
+      "dynamic-client-registration": false,
+      "business-to-consumer-tax-calculation-net": false,
+      "business-to-business-tax-calculation-net": false
+    }
+  },
+  "sellers": {
+    "primary": {
```
lukehesluke, line 34: Just a suggestion but I wonder if there needs to be either or both of the following:

* Assertion that if primary authentication style is client credentials then 1). there are no other sellers configured 2). the multiple-sellers features are disabled. Maybe this could be a test
* Documentation of Client Credentials mode in configuration which informs of this constraint

packages/openactive-openid-test-client/lib/auth-key-manager.js
```diff
@@ -138,14 +139,41 @@ module.exports = class OpenActiveTestAuthKeyManager {
...
+        const { tokenSet } = await this.client.authorizeClientCredentialsFlow(clientId, clientSecret, this.client.issuer, false);
         this.bookingPartnersConfig[bookingPartnerIdentifier].authentication.orderFeedTokenSet = tokenSet;
         this.log(`Retrieved Orders Feed tokens via Client Credentials Flow for booking partner '${bookingPartnerIdentifier}'`);
       } catch (error) {
         this.log(`Error retrieving Orders Feed tokens via Client Credentials Flow for booking partner '${bookingPartnerIdentifier}'`);
         throw error;
       }
     }));
+
+    // Refresh Seller token if Single Seller booking system
```
lukehesluke, line 151:
```suggestion
    // Refresh Seller token if Client Credentials Auth Flow is used.
    // We only consider the primary seller as Client Credentials only supports 1 seller.
```

packages/openactive-openid-test-client/lib/auth-key-manager.js
```diff
@@ -138,14 +139,41 @@ module.exports = class OpenActiveTestAuthKeyManager {
...
+    if (this.sellersConfig.primary?.authentication?.clientCredentials) {
+      // If Booking System is a single seller system, add the tokenSet to the primary (and only) Seller
```
lukehesluke, line 153:
```suggestion
      // Since Booking System is a single seller system, add the tokenSet to the primary (and only) Seller
```

?

lukehesluke (changes requested):

config/.example.single-seller-client-credentials.json
civsiv: Thanks @lukehesluke.
I have updated the broker documentation for Client Credentials, and have raised an issue here: https://github.com/openactive/openactive-test-suite/issues/520

PR #518, opened 2023-01-31
fix: add openBookingInAdvance constraint to TestOpportunityBookableOu…

civsiv:
…tsideValidFromBeforeStartDate

For this test, the offer must be theoretically bookable but currently outside the booking window. This test was picking offers that were not theoretically bookable (openBookingInAdvance: https://openactive.io/Unavailable)


lukehesluke (approved):
Good catch

PR #519, opened 2023-02-02
Feature/single multiple seller config test

civsiv:
Addresses https://github.com/openactive/openactive-test-suite/issues/520


lukehesluke:
I believe this is now obsolete as it was replaced with this PR: https://github.com/openactive/openactive-test-suite/pull/640

Apologies I had not seen this PR before I did mine 😞 but this one would require quite a bit of updating anyway

PR #521, opened 2023-02-06
feat: add openBookingInAdvance filter to test internal criteria

civsiv:


packages/test-interface-criteria/src/criteria/internal/InternalCriteriaFutureScheduledOpportunity.js
```diff
@@ -38,6 +39,10 @@ const InternalCriteriaFutureScheduledOpportunity = createCriteria({
       'Offer must not be non-free with openBookingPrepayment unavailable',
       excludePaidBookableOffersWithPrepaymentUnavailable,
     ],
+    [
+      'openBookingInAdvance of offer must not be `https://openactive.io/Unavailable`',
+      mustNotBeOpenBookingInAdvanceUnavailable,
+    ],
   ],
   testDataShape: (options) => ({
     opportunityConstraints: ({
```
lukehesluke, line 48: new constraints should be added here

packages/test-interface-criteria/src/criteria/internal/InternalCriteriaFutureScheduledOpportunity.js
```diff
@@ -38,6 +39,10 @@ const InternalCriteriaFutureScheduledOpportunity = createCriteria({
       'Offer must not be non-free with openBookingPrepayment unavailable',
       excludePaidBookableOffersWithPrepaymentUnavailable,
     ],
+    [
```
lukehesluke, line 42: I'm having a hard time judging whether this constraint makes sense in `InternalCriteriaFutureScheduledOpportunity`. Just based on the name itself, I take it that this criteria refers to something that is scheduled in the future. For which, the 2 opportunity constraints already present make sense. But the existing offer constraint and this new constraint add a new dimension to that criteria which is that the opportunity must be Open Bookable.

So I think it would be worth doing one of the following things:

1. Create a separate constraint called `InternalCritieraFutureScheduledAndBookable` (or something)
2. If it looks like all constraints and tests which use this constraint also all require bookability, maybe just change this constraint's name to the above
3. Or separately add the bookability constraints to the relevant criteria but remove them from this internal criteria

lukehesluke (changes requested):

civsiv:
PR closed to due to:
1) There's a better way to do this (`shapeConstraintRecipes.mustHaveBookableOffer(options)`)
2) This is not the best place to do it

PR #522, opened 2023-02-06
feat: add bookable offer constraint to BookableNoSpaces criteria

civsiv:


packages/test-interface-criteria/src/criteria/TestOpportunityBookableNoSpaces.js
```diff
@@ -25,13 +25,16 @@ const TestOpportunityBookableNoSpaces = createCriteria({
     ],
   ],
   offerConstraints: [],
```
lukehesluke, line 27: Does it also need to be here at the moment or are these fields now obsolete?

I remember the idea was to move everything to the testDataShape approach but that might be something that never quite got finished...

I'll have a quick look

lukehesluke: Sorry @civsiv it's also got to have these:

https://github.com/openactive/openactive-test-suite/blob/4865a9a3d146e7b34d58ee86eb84fad5c0850870/packages/test-interface-criteria/src/criteria/TestOpportunityBookable.js#L28-L37

This is because `offerConstraints` are still used for most criteria filtering. Although would be interesting to know if it passed on your local machine without it 🤔

Issue #534, opened 2023-04-11
It is not possible to pass Test Suite unless Seller Requested Cancellation is enabled


lukehesluke:
The [orders-updated-then-deleted](https://github.com/openactive/openactive-test-suite/blob/master/packages/openactive-integration-tests/test/features/core/order-deletion/implemented/orders-updated-then-deleted-test.js) test is part of the [core/order-deletion](https://github.com/openactive/openactive-test-suite/blob/master/packages/openactive-integration-tests/test/features/core/order-deletion/README.md) feature. This is a required feature and yet it uses a simulated Seller Cancellation in order to achieve its means. But, [seller-requested-cancellation](https://github.com/openactive/openactive-test-suite/blob/master/packages/openactive-integration-tests/test/features/cancellation/seller-requested-cancellation/README.md) is an optional feature.

One very easy way to fix this would be to:

- only run the test if seller-requested-cancellation is enabled. I believe this is easily achievable by using the `runOnlyIf` arg to `FeatureHelpers.describeFeature(..)`. Example usage: [here](https://github.com/openactive/openactive-test-suite/blob/master/packages/openactive-integration-tests/test/features/payment/payment-reconciliation-detail-validation/not-implemented/payment-reconciliation-detail-validation-test.js#L15)

lukehesluke:
It's definitely worth debating if there's some way to test the `orders-updated-then-deleted` test without seller-requested-cancellation being required. This would increase the test coverage. However, because this issue actually makes it impossible to run Test Suite presently without seller-requested-cancellation, I believe it may be worth merging the following PR even though it just implements the "easy fix" mentioned in this issue's description:

- PR: https://github.com/openactive/openactive-test-suite/pull/537

lukehesluke:
Closed due to completion in this PR: https://github.com/openactive/openactive-test-suite/pull/537

PR #537, opened 2023-05-10
fix: Only run orders-updated-then-deleted if seller-requested-cancellation is implemented

lukehesluke:
See issue: https://github.com/openactive/openactive-test-suite/issues/534


lukehesluke:
[LTA - Test Suite pass for MS2](https://trello.com/c/1JVOT1wS/4553-lta-test-suite-pass-for-ms2)

nickevansuk:
Should be updated to "only if seller-requested-cancellation or customer-requested-cancellation is implemented", as an Orders feed implementation is only required to support either of these (aside from approval flow)

nickevansuk (approved):
Reflecting on my comment above, it is very unlikely that `customer-requested-cancellation` is implemented without `seller-requested-cancellation`, therefore this comment should be implemented as a nice-to-have

PR #546, opened 2023-07-07
feat: get broker booking partner identifiers from config

civsiv:
Currently the Booking Partner identifiers are hardcoded to be `primary` and `secondary` regardless of if these are set in config. When there was no `secondary` booking partner config set, this results in requests hitting the maximum retries and the app crashing. 
This fix gets the booking partner identifiers from config and only uses the ones set there.

Before
Screenshot 2023-07-07 at 17 22 06

Now:
Screenshot 2023-07-07 at 17 18 04

### DIFF
--- a/packages/openactive-broker-microservice/README.md
+++ b/packages/openactive-broker-microservice/README.md
@@ -158,7 +158,7 @@ While debugging authentication it can be useful to log the configuration that th
 
 Config for [Booking Partners](https://openactive.io/open-booking-api/EditorsDraft/1.0CR3/#dfn-booking-partner) that the Test Suite will use to connect to your Booking System.
 
-Broker uses two Booking Partners, named `primary` and `secondary`.
+When running Broker Microservice with [Integration Tests](../openactive-integration-tests/) (which is the primary usage of Broker Microservice), you muse define two Booking Partners, named `primary` and `secondary`.
 
 ```json
   "bookingPartners": {
@@ -177,6 +177,8 @@ Broker uses two Booking Partners, named `primary` and `secondary`.
     }
   }
 ```
+
+However, when running Broker Microservice as a standalone app, you may define any number of Booking Partners, with whatever identifiers you like e.g. `acme1`, `acme2` and `acme3` instead of `primary` and `secondary`. Just know that these will NOT work with the [Integration Tests](../openactive-integration-tests/), which expect only `primary` and `secondary`.
 
 The `authentication` field can contain one of several different authentication strategies for authenticating as the Booking Partner for Order and Order Proposal RPDE Feed requests. Set only one of these strategies. The different authentication strategies are documented in the below subsections.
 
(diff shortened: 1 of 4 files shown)


lukehesluke (changes requested):
## Integration Test compatability

This PR allows for any possible Booking Partner identifiers to be used in Broker Microservice, (e.g. `partnerOne`, `partnerTwo`, `partnerThree`) which is great.

However, these would then not work when running integration tests as there's more hardcoded logic in there.

So this PR definitely has merit to merge in as-is because Broker Microservice can be used as a standalone service, and this PR improves the functionality in that use case. But, it allows for a behaviour which will definitely break Integration Tests. Maybe if you could document that this is the case?

_Edit: This has been done ✅_

## Orders Feed won't monitor

Also, app.js:1680 would need fixing in order to :

>           processPage: feedBookingPartnerIdentifier === 'primary'

It looks like this means that only an Orders Feeds with identifier `primary` will be actually monitored. Not sure why this logic is in there tbh?

_Edit: Yes, according to the TODO in this line from an earlier commit, this is just an accident: https://github.com/openactive/openactive-test-suite/blame/c4260689ff9c71c066d5d9d6d719f0af7c886afc/packages/openactive-broker-microservice/app.js#L1634. So we can now fix this_

_Edit: This has been done ✅_

## Docs

the `bookingPartner` docs in broker's README needs updating. Perhaps this could include the bit about Integration Tests compatability

_Edit: This has been done ✅_

## Misc

packages/openactive-broker-microservice/src/models/core.d.ts:5. Type needs updating

_Edit: This has been done ✅_

packages/openactive-broker-microservice/README.md
```diff
@@ -178,6 +178,8 @@ Broker uses two Booking Partners, named `primary` and `secondary`.
   }
 ```
 
+However, when running Broker Microservice as a standalone app, you may define any number of Booking Partners, with whatever identifiers you like e.g. `acme1`, `acme2` and `acme3` instead of `primary` and `secondary`. Just know that these will NOT work with the [Integration Tests](../openactive-integration-tests/), which expect only `primary` and `secondary`.
```
nickevansuk, line 181: Consistent impersonal tone with rest of doc

```suggestion
When running Broker Microservice as a standalone app, any number of Booking Partners may be defined e.g. `acme1`, `acme2` and `acme3` instead of `primary` and `secondary`. Note that these will _not_ work with the [Integration Tests](../openactive-integration-tests/), which expect only `primary` and `secondary`.
```

nickevansuk (approved):
Approved with one doc change

lukehesluke:
[Complete LTA documentation for Customer Requested Cancellation and other Orders Feed-related processes](https://trello.com/c/PYv6Cne8/4664-complete-lta-documentation-for-customer-requested-cancellation-and-other-orders-feed-related-processes)

PR #552, opened 2023-08-04
feat: Idempotency support

nickevansuk:
Reintroduce commented out idempotency testing, in parallel to its support within reference implementation (https://github.com/openactive/OpenActive.Server.NET/issues/100, https://github.com/openactive/OpenActive.Server.NET/pull/207).

Note these tests are failing due to the following issues:
1. `idempotentRepeatBAfterBook` needs to be wired up to work with `initialiseSimpleC1C2BookFlow`
2. `payment.identifier` is currently randomly generated for each B request, which makes the request body unique. `payment.identifier` needs to be consistent across the two B calls in order for the second call to trigger the idempotency logic, which is based on two _identical_ requests.

The reference implementation support for idempotency has been tested independently of the above and seems to work as expected.


packages/openactive-integration-tests/test/features/payment/free-opportunities/implemented/opportunity-free-idempotency-test.js
```diff
@@ -0,0 +1,33 @@
...
+    fetchOpportunities,
+    bookRecipe,
+    defaultFlowStageParams,
+    bookRecipeArgs,
+  } = FlowStageRecipes.initialiseSimpleC1C2BookFlow(orderItemCriteriaList, logger);
+  const idempotentRepeatB = FlowStageRecipes.idempotentRepeatBAfterBook(
+    orderItemCriteriaList,
+    bookRecipe,
+    defaultFlowStageParams,
+    omit(bookRecipeArgs, ['prerequisite']),
```
lukehesluke, line 26: One issue was this: Idempotent B wasn't getting the same starting args as B, which was messing up its expectations for what capacity should be asserted for after the 2nd B

packages/openactive-integration-tests/test/helpers/flow-stages/b.js
```diff
@@ -52,6 +67,7 @@ async function runB({ templateRef, accessPass, brokerRole, uuid, sellerConfig, c
     brokerRole,
     positionOrderIntakeFormMap,
     customer,
+    paymentIdentifier: paymentIdentifierIfPaid,
```
lukehesluke, line 70: The other issue was, as you mentioned, Nick, that `paymentIdentifier` was being re-generated for each B/P call. Rather than using stage input/output, I've just put it as an arg to the B / P stages, which should make it very easy to do things like test what happens if the repeated B call has a different payment identifier, if a B-after-P call has a different payment identifier, etc

packages/openactive-integration-tests/test/templates/common.js
```diff
@@ -18,21 +17,22 @@ const shortid = require('shortid');
  */
 
 /**
- * @param {boolean} [includeIdentifier]
+ * @param {string | null | undefined} [identifier]
```
nickevansuk, line 20: Do we need to add more detail to this comment to explain the difference between passing `null` and `undefined` here?

nickevansuk (commented):
One small comment on this

PR #553, opened 2023-08-11
feat: Docker with package upgrades

nickevansuk:


.github/workflows/docker.yml
```diff
@@ -0,0 +1,49 @@
...
+          tags: |
+            latest
+
+      - name: Build and push Docker image for OpenActive Test Suite
+        uses: docker/build-push-action@v4
+        with:
+          context: .
+          push: true
+          tags: ${{ steps.meta.outputs.tags }}
+          labels: ${{ steps.meta.outputs.labels }}
```
nickevansuk, line 49: This needs to be expanded to test before push (https://docs.docker.com/build/ci/github-actions/test-before-push/), to ensure that everything works as expected.

The reference implementation service (https://github.com/openactive/OpenActive.Server.NET/blob/d6e2f57dd8b6a7835cf8bcd747eef7553a7128f4/.github/workflows/docker-test.yaml#L16-L32) can be added here to facilitate this.

The config can also be passed in (https://github.com/openactive/OpenActive.Server.NET/blob/d6e2f57dd8b6a7835cf8bcd747eef7553a7128f4/.github/workflows/docker-test.yaml#L53-L56)

Though note that we'd need to get [ghcr.io/openactive/reference.bookingsystem.identity](http://ghcr.io/openactive/reference.bookingsystem.identity) running to actually test Puppeteer

lukehesluke:
Replaced by:

* https://github.com/openactive/openactive-test-suite/pull/596
* AND https://github.com/openactive/openactive-test-suite/pull/597

Issue #595, opened 2023-09-21
Upgrade Test Suite Puppeteer in preperation for Docker


lukehesluke:
Upgrade puppeteer to the latest version. This enables dockerisation of Test Suite (issue coming soon), which is in necessary as publishers often have challenges running Test Suite on their local machine and in their CI environments.

The OpenID test Client package has also been split to resolve compatibility issues between Puppeteer and Jest, both now and in the future. This has a knock-on impact on the documentation around and relating to this repo.

PR #596, opened 2023-09-21
chore: Upgrade Puppeteer and node version; split OpenID test Client package

lukehesluke:
Closes https://github.com/openactive/openactive-test-suite/issues/595

Also:

* Upgrade node version, which was required for puppeteer
* Split up the openid-test-client package into 3 packages, OpenID Client, OpenID Browser Automation and OpenID Test CLI

    This is because Puppeteer has some clashes with Jest. By splitting these packages up, Puppeteer is only required in the OpenID Browser Automation package. This package is then only used by Broker Microservice, which does not run jest tests. An additional side effect from this is that Puppeteer, which is a very large package, doesn't leak out into any packages where it's not needed
* Fix the OpenID Test CLI
    * Closes https://github.com/openactive/openactive-test-suite/issues/599 
* Give summaries of each of the 3 packages:
    * Closes https://github.com/openactive/openactive-test-suite/issues/600


packages/openactive-openid-test-cli/lib/cli.js
```diff
@@ -83,29 +84,26 @@ const argv = yargs(process.argv.slice(2))
       type: 'string',
       alias: 'u',
       description: 'Username of the Seller',
-      default: 'test',
+      default: 'test1',
```
lukehesluke, line 87: This matches the details for the primary seller in Ref Impl

packages/openactive-openid-test-cli/lib/cli.js
```diff
@@ -83,29 +84,26 @@ const argv = yargs(process.argv.slice(2))
...
     password: {
       type: 'string',
       alias: 'p',
       description: 'Password of the Seller',
-      default: 'test',
+      default: 'test1',
     },
     usernameFieldSelector: {
       type: 'string',
-      alias: 'l',
```
lukehesluke, line 96: These all used the same alias and this caused all of the selectors to be `[name='username' i]`, which lead to CLI failing

nickevansuk (approved):
Approved noting that CI is presently failing

PR #604, opened 2023-10-02
feat: Add `validThroughBeforeStartDate`

nickevansuk:
Add basic support for `validThroughBeforeStartDate` to ensure that test suite does not fail for systems that implement this (as per #605)

### DIFF
--- a/packages/test-interface-criteria/built-types/types/Criteria.d.ts
+++ b/packages/test-interface-criteria/built-types/types/Criteria.d.ts
@@ -5,7 +5,7 @@ import { TestDataShape } from './TestDataShape';
 
 export type OpportunityConstraint = (opportunity: Opportunity, options?: Options) => boolean;
 
-export type OfferConstraint = (offer: Offer, opportunity: Opportunity, options?: Options) => boolean;
+export type OfferConstraint = (offer: Offer, opportunity?: Opportunity, options?: Options) => boolean;
 
 export type TestDataShapeFactory = (options: Options) => TestDataShape;
 
--- a/packages/test-interface-criteria/built-types/types/Criteria.d.ts
+++ b/packages/test-interface-criteria/built-types/types/Criteria.d.ts
@@ -5,7 +5,7 @@ import { TestDataShape } from './TestDataShape';
 
 export type OpportunityConstraint = (opportunity: Opportunity, options?: Options) => boolean;
 
-export type OfferConstraint = (offer: Offer, opportunity?: Opportunity, options?: Options) => boolean;
+export type OfferConstraint = (offer: Offer, opportunity: Opportunity, options?: Options) => boolean;
 
 export type TestDataShapeFactory = (options: Options) => TestDataShape;
 
--- a/packages/test-interface-criteria/src/criteria/TestOpportunityBookableOutsideValidFromBeforeStartDate.js
+++ b/packages/test-interface-criteria/src/criteria/TestOpportunityBookableOutsideValidFromBeforeStartDate.js
@@ -18,8 +18,8 @@ function mustHaveBookingWindowAndBeOutsideOfIt(offer, opportunity, options) {
   /* If, within 2 hours, the booking window would be reached, it may be possible for this to happen
   during the test run. So, to be on the safe side, we only accept Opportunities whose booking window
   starts at least 2 hours in the future. */
-  return (dateAfterWhichBookingsCanBeMade == null || options.harvestStartTimeTwoHoursLater < dateAfterWhichBookingsCanBeMade)
-    && (dateBeforeWhichBookingsCanBeMade == null || options.harvestStartTime > dateBeforeWhichBookingsCanBeMade);
+  return (dateAfterWhichBookingsCanBeMade !== null && options.harvestStartTimeTwoHoursLater < dateAfterWhichBookingsCanBeMade)
+    || (dateBeforeWhichBookingsCanBeMade !== null && options.harvestStartTime > dateBeforeWhichBookingsCanBeMade);
 }
 
 /**
--- a/packages/test-interface-criteria/src/criteria/TestOpportunityBookableOutsideValidFromBeforeStartDate.js
+++ b/packages/test-interface-criteria/src/criteria/TestOpportunityBookableOutsideValidFromBeforeStartDate.js
@@ -19,7 +19,7 @@ function mustHaveBookingWindowAndBeOutsideOfIt(offer, opportunity, options) {
   during the test run. So, to be on the safe side, we only accept Opportunities whose booking window
   starts at least 2 hours in the future. */
   return (dateAfterWhichBookingsCanBeMade == null || options.harvestStartTimeTwoHoursLater < dateAfterWhichBookingsCanBeMade)
-    || (dateBeforeWhichBookingsCanBeMade == null || options.harvestStartTime > dateBeforeWhichBookingsCanBeMade);
+    && (dateBeforeWhichBookingsCanBeMade == null || options.harvestStartTime > dateBeforeWhichBookingsCanBeMade);
 }
 
 /**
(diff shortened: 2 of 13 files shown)


packages/test-interface-criteria/src/criteria/TestOpportunityBookableOutsideValidFromBeforeStartDate.js
```diff
@@ -11,13 +11,15 @@ const { InternalCriteriaFutureScheduledAndDoesNotRequireDetails } = require('./i
...
+  const dateBeforeWhichBookingsCanBeMade = getDateBeforeWhichBookingsCanBeMade(offer, opportunity);
+  if (dateAfterWhichBookingsCanBeMade == null && dateBeforeWhichBookingsCanBeMade == null) {
     return false; // has no booking window
   }
   /* If, within 2 hours, the booking window would be reached, it may be possible for this to happen
   during the test run. So, to be on the safe side, we only accept Opportunities whose booking window
   starts at least 2 hours in the future. */
-  return options.harvestStartTimeTwoHoursLater < dateAfterWhichBookingsCanBeMade;
+  return (dateAfterWhichBookingsCanBeMade == null || options.harvestStartTimeTwoHoursLater < dateAfterWhichBookingsCanBeMade)
+    || (dateBeforeWhichBookingsCanBeMade == null || options.harvestStartTime > dateBeforeWhichBookingsCanBeMade);
```
civsiv, line 22:
```suggestion
    || (dateBeforeWhichBookingsCanBeMade == null || options.harvestStartTimeTwoHoursLater > dateBeforeWhichBookingsCanBeMade);
```

Want to be outside the 2 hour window right? The criteria below indicates that too. But timezones mess with my head so I could have read this wrong.

civsiv (approved):

packages/test-interface-criteria/src/criteria/criteriaUtils.js
```diff
@@ -376,8 +391,13 @@ function mustNotBeOpenBookingInAdvanceUnavailable(offer) {
 */
 function mustHaveBeInsideValidFromBeforeStartDateWindow(offer, opportunity, options) {
```
lukehesluke, line 392: This function name is confusing. Could now be called `mustBeInsideBookingWindowIfOneExists(..)` to mirror `mustHaveBookingWindowAndBeOutsideOfIt(..)`

packages/test-interface-criteria/src/criteria/criteriaUtils.js
```diff
@@ -376,8 +391,13 @@ function mustNotBeOpenBookingInAdvanceUnavailable(offer) {
...
   const dateAfterWhichBookingsCanBeMade = getDateAfterWhichBookingsCanBeMade(offer, opportunity);
-  if (dateAfterWhichBookingsCanBeMade == null) { return true; } // no booking window - therefore bookable at any time
-  return options.harvestStartTime > dateAfterWhichBookingsCanBeMade;
+  const dateBeforeWhichBookingsCanBeMade = getDateBeforeWhichBookingsCanBeMade(offer, opportunity);
+  if (dateAfterWhichBookingsCanBeMade == null && dateBeforeWhichBookingsCanBeMade == null) { return true; } // no booking window - therefore bookable at any time
```
lukehesluke, line 395: This line can be removed

packages/test-interface-criteria/src/criteria/criteriaUtils.js
```diff
@@ -287,6 +287,21 @@ function getDateAfterWhichBookingsCanBeMade(offer, opportunity) {
   return dateMinusDuration(opportunity.startDate, offer.validFromBeforeStartDate);
 }
 
+/**
+ * Get the date that the startDate - validThroughBeforeStartDate window starts
+ *
+ * @param {Offer} offer
+ * @param {Opportunity} opportunity
+ * @returns {DateTime | null} null if there is no booking window defined.
```
lukehesluke, line 295:
```suggestion
 * @returns {DateTime | null} null if there is no booking window upper limit defined.
```

lukehesluke: and then likewise for the above function

packages/test-interface-criteria/src/criteria/TestOpportunityBookableOutsideValidFromBeforeStartDate.js
lukehesluke: I think this is right as is. If it was outside the "2 hours later" time, then this criteria could falsely identify an event that is bookable until 1h30-after-harvest-start as not-bookable-due-to-being-outside-of-the-booking-window. Then, if an attempt is made to book the event 30mins into harvest start, it would succeed, which goes against the criteria

packages/test-interface-criteria/src/criteria/TestOpportunityBookableOutsideValidFromBeforeStartDate.js
```diff
@@ -51,6 +53,9 @@ const TestOpportunityBookableOutsideValidFromBeforeStartDate = createCriteria({
       'oa:validFromBeforeStartDate': dateRange({
         minDate: options.harvestStartTimeTwoHoursLater.toISO(),
       }),
+      'oa:validThroughBeforeStartDate': dateRange({
+        maxDate: options.harvestStartTimeTwoHoursLater.toISO(),
```
lukehesluke, line 57: options.harvestStartTime

lukehesluke (changes requested):

packages/test-interface-criteria/src/criteria/criteriaUtils.js
nickevansuk: Great point, have done so

packages/test-interface-criteria/src/criteria/criteriaUtils.js
nickevansuk: Good idea, have clarified

packages/test-interface-criteria/src/criteria/criteriaUtils.js
nickevansuk: Good spot

packages/test-interface-criteria/src/criteria/TestOpportunityBookableOutsideValidFromBeforeStartDate.js
nickevansuk: Agree, it looks right as is - good to check though!

packages/test-interface-criteria/src/criteria/TestOpportunityBookableOutsideValidFromBeforeStartDate.js
nickevansuk: Good spot!

packages/test-interface-criteria/built-types/types/Criteria.d.ts
```diff
@@ -5,7 +5,7 @@ import { TestDataShape } from './TestDataShape';
 
 export type OpportunityConstraint = (opportunity: Opportunity, options?: Options) => boolean;
 
-export type OfferConstraint = (offer: Offer, opportunity: Opportunity, options?: Options) => boolean;
+export type OfferConstraint = (offer: Offer, opportunity?: Opportunity, options?: Options) => boolean;
```
nickevansuk, line 8: @lukehesluke this change seems to create quite significant knock-on effects within the generated types. It looked like it was a typo that needed fixing, but I'm wondering if it was actually intentional?

Would you be able to shed any light on this?

lukehesluke: @nickevansuk why did this seem like a typo? This was made quite a while ago but it probably was intentional

nickevansuk: For the record: this was a result of upgrading TS locally. Reverting TS removed the need to make any type changes at all.

PR #614, opened 2023-11-09
feat: Randomly exclude optional customer fields

### LINKED ISSUE #613: Ensure that optional customer fields are not required by the Booking System under test
In the [Open Booking API spec](https://openactive.io/open-booking-api/EditorsDraft/1.0CR3/#schema-person-for-customer-or-attendee), `email` is the only field that is required for a customer.

We should ensure that the Booking System under test does not require any other data for the customer.

I've seen an instance in which a Booking System passed Test Suite but would break if the Broker does not pass `.customer.telephone` to C2 or B.

Two possible approaches: 

## 1. Have FlowStageUtils.createRandomCustomerDetails(...) not always create all fields

Randomise whether or not `FlowStageUtils.createRandomCustomerDetails(...)` always returns all fields. It should always return type and email but only optionally the other fields

## 2. Extra tests

Extra tests to ensure more strictly that a passing Booking system explicitly supports all fields being set and no fields being set.

----

Personally, I would prefer approach # 1. I think it's an extension of the fact that tests already include random data. It has the disadvantage of being able to (very rarely) create false negatives where a Booking System passes all tests but actually cannot support some particular use case like `telephone` not existing. However, if we apply the same ethos to all data types, this could require a huge number of tests.

lukehesluke:
Closes #613

### DIFF
--- a/packages/openactive-integration-tests/test/features/details-capture/customer-details-capture-non-essential/implemented/non-essential-customer-details-reflected-test.js
+++ b/packages/openactive-integration-tests/test/features/details-capture/customer-details-capture-non-essential/implemented/non-essential-customer-details-reflected-test.js
@@ -16,7 +16,9 @@ FeatureHelper.describeFeature(module, {
 },
 function (configuration, orderItemCriteriaList, featureIsImplemented, logger) {
   // # Initialise Flow Stages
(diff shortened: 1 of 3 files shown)


lukehesluke:
[Test Suite: Randomly exclude optional customer deets (if quick)](https://trello.com/c/K0rm2mgl/4795-test-suite-randomly-exclude-optional-customer-deets-if-quick)

nickevansuk (approved):
Great idea

Issue #520, opened 2023-02-02
[<Feature Name>] - [<Test Identifier>] - <Use Case>


civsiv:
**Feature**

core/single-seller
core/multiple-sellers

**Use Case**

Single-Seller and Multiple-Sellers are mutually exclusive and one feature has to be implemented. Write tests that make sure that:
1) If one is implemented, the other is explicitly not implemented
2) If Single-Seller is implemented, SELLER_CONFIG only has the `primary` seller.

lukehesluke:
@civsiv possible additional test case:

3. If Multi-Seller is implemented. None of the sellers in SELLER_CONFIG can include `authentication.clientCredentials`

lukehesluke:
Was reminded once more that this was needed. +1!

Otherwise a Booking System with multiple Sellers could just not turn on the `multiple-sellers` feature when running tests (e.g. because they didn't notice it) and they would not get this very important bit of test coverage

lukehesluke:
> If Multi-Seller is implemented. None of the sellers in SELLER_CONFIG can include authentication.clientCredentials

I've decided not to do this because it does seem plausible that multiple sellers could be supported while also supporting the `authentication.clientCredentials` auth strategy e.g. there are multiple sellers but they don't have separate auth

lukehesluke:
Also spawned this issue to constrain this even further: https://github.com/openactive/openactive-test-suite/issues/641

PR #640, opened 2024-02-21
feat: Add single-seller feature, which is mutually exclusive with multiple-sellers

lukehesluke:
Closes https://github.com/openactive/openactive-test-suite/issues/520


packages/openactive-integration-tests/test/features/payment/prepayment-required-unavailable/not-implemented/prepayment-required-unavailable-not-in-use-test.js
```diff
@@ -1,6 +1,6 @@
 const { FeatureHelper } = require('../../../../helpers/feature-helper');
 
-FeatureHelper.describeFeatureShouldBeImplementedIfOtherFeaturesAre(module, {
+FeatureHelper.describeFeatureShouldBeImplementedIfOtherFeaturesAreOrAreNot(module, {
```
civsiv, line 3: I find this method name really confusing but don't really have any suggestions for an alternative. Could this be split into two methods? 
`describeFeatureShouldBeImplementedIfOtherFeaturesAre`
`describeFeatureShouldBeImplementedIfOtherFeaturesAreNot`

I can't see any examples that use both (but I understand that it could in theory).

packages/openactive-integration-tests/documentation/generator.js
```diff
@@ -77,6 +77,15 @@ const INDEX_TESTS_IMPLEMENTED_JSON_FILE = path.join(FEATURES_ROOT, 'tests-implem
  * }} FeatureMetadataItem
  */
 
+const FEATURES_NOT_REQUIRED_FOR_DEFAULT_JSON = new Set([
+  'customer-requested-cancellation-always-allowed',
```
civsiv, line 81: Is the reason this is not in the default because customer cancellation are not always allowed? If so it's a weird feature but ok 🤷🏽

civsiv (approved):

packages/openactive-integration-tests/test/features/payment/prepayment-required-unavailable/not-implemented/prepayment-required-unavailable-not-in-use-test.js
lukehesluke: Yes you're totally right. I've done that and that makes it easier to understand. And even if there were examples that used both, they can easily just be two separate tests

PR #655, opened 2024-03-13
feat(tests): Include required Test Interface Actions in generated test docs

### LINKED ISSUE #654: Test Suite: Have automatic documentation generation include whether or not a Test Interface action is required
It's currently left to the Test Suite user to read through the Test Interface spec and figure out which actions they need to implement based on the features that they implement.

I suspect this mostly happens either with guidance from an expert or by running tests and then seeing an error that the required action is not implemented (and then likely being confused)

It would be good for them to have this knowledge upfront so that they can plan accordingly and give accurate estimates, etc.

----

This might be done by adding a param to FeatureHelper. Bonus points if FeatureHelper can somehow assert that a test only uses and uses all actions that have been configured

lukehesluke:
Closes #654 

What this does:

1. Introduce a new `FeatureHelper.describeFeature(..)` param to annotate which TI Actions are used in a test
2. `FeatureHelper.describeFeature(..)` will now, via `DescribeFeatureRecord`, check, after a test run completes, that the TI Actions used are the same as annotated
    a. Test report generation only renders this check if it's the only thing that fails. And it will include a special message to ensure that this issue gets looked at by maintainers. The rationale for all this can be seen in code comments
3. The TI Actions are included in the auto-generated documentation for each test, and the summary documentation across all tests

## To the Reviewer

This is a large PR, but a lot of this is just changes to the auto-generated docs (one for each feature!) and piping in a new parameter in all the various `-test.js` files, so I think you can skim through the whole `packages/openactive-integration-tests/test/features/` folder. The core of the work is in the other files

## QA

**QA of the 3. behaviour (doc generation)**

Single feature:

Screenshot 2024-03-13 at 13 24 23

Summary:

Screenshot 2024-03-13 at 13 24 41

(one issue is that the table is now quite wide, but given the time spent so far, perhaps this could be turned into an issue for the future)

**QA of the 2.a. behaviour (in report generation)**

If the test otherwise passes, but the annotation was incorrect, that check is shown:

Screenshot 2024-03-13 at 12 51 24

If the test otherwise fails, the check is hidden:

Screenshot 2024-03-13 at 12 53 22

If the test succeeds and the annotation was correct, the check is hidden:

Screenshot 2024-03-13 at 12 58 10


packages/openactive-integration-tests/test/helpers/flow-stages/flow-stage-utils.js
```diff
@@ -157,6 +159,7 @@ const FlowStageUtils = {
       sellerConfig: sellerConfig || SELLER_CONFIG.primary,
       customer: this.createRandomCustomerDetails(includeAllOptionalCustomerDetails ?? false),
       orderItemCriteriaList,
+      describeFeatureRecord,
```
lukehesluke, line 162: This will now be input into all the FlowStage constructors. It's used only by the Test Interface Action Flow Stage for now

packages/openactive-integration-tests/test/helpers/flow-stages/test-interface-action.js
```diff
@@ -39,10 +39,12 @@ const { FlowStageUtils } = require('./flow-stage-utils');
  * @param {object} args
  * @param {ActionSpec} args.action
  * @param {RequestHelperType} args.requestHelper
+ * @param {import('../describe-feature-record').DescribeFeatureRecord} args.describeFeatureRecord
  * @returns {Promise<Output>}
  */
-async function runTestInterfaceAction({ action, requestHelper }) {
+async function runTestInterfaceAction({ action, requestHelper, describeFeatureRecord }) {
   const response = await requestHelper.callTestInterfaceAction({ action });
+  describeFeatureRecord.recordTestInterfaceActionUse(action.type);
```
lukehesluke, line 47: This is how TI Action use is tracked over a test run

packages/openactive-integration-tests/test/helpers/logger.js
```diff
@@ -56,6 +64,56 @@ const { getConfigVarOrThrow } = require('./config-utils');
  *   };
  *   events: FlowStageLogEvent[]
  * }} FlowStageLog
+ *
+ * @typedef {{
```
lukehesluke, line 68: I also typed some things as I was trying to understand how it all worked

packages/openactive-integration-tests/test/reporter.js
```diff
@@ -1,3 +1,11 @@
+/**
+ * A custom Jest reporter
+ *
+ * NOTE: Because reporter.js is loaded in a separate Node.js process (by Jest),
+ * it does not have access to the same environment as the tests. One
+ * consequence of this is that it cannot import modules that use any of the
+ * Jest globals like `afterEach`, etc.
```
lukehesluke, line 7: This happened to me and was quite confusing!

packages/openactive-integration-tests/test/helpers/feature-helper.js
```diff
@@ -169,6 +185,50 @@ class FeatureHelper {
...
+            ) {
+              return allUsedTestInterfaceActions
+                .filter(x => x !== 'test:SellerAcceptOrderProposalSimulateAction');
+            }
+            return allUsedTestInterfaceActions;
+          })();
+
+          /* JSON.stringify(..) used because otherwise the test output truncates
+          to `expected [Array(2)] to equal [Array(1)]`, which is useless. */
+          expect(JSON.stringify(usedTestInterfaceActions))
```
civsiv, line 226:
```suggestion
          /* JSON.stringify(..) used because otherwise the test output truncates
          to `expected [Array(2)] to equal [Array(1)]`, which is useless. 
          The arrays are sorted so string comparison is fine. */
          expect(JSON.stringify(usedTestInterfaceActions))
```
My comment might not be needed 🤷🏽‍♂️

civsiv (approved):

PR #658, opened 2024-03-14
feature: fix validation-errors.handlebars path

civsiv:
Before:
Screenshot 2024-03-14 at 12 34 59

After:
Screenshot 2024-03-14 at 12 35 15


lukehesluke (approved):
Great find!

PR #663, opened 2024-03-19
feat: add IS_CI env var to Ref Impl workflow

civsiv:


civsiv:
This functionality is no longer needed as Reference Implementation now runs in CI mode by default

Issue #669, opened 2024-03-25
Broker: Consolidate opportunityItemRowCache and opportunityCache to reduce memory usage & simplify code


lukehesluke:
Broker uses two strategies to cache full opportunity data: opportunityItemRowCache and opportunityCache (see their documentation in packages/openactive-broker-microservice/src/state.js).

In both cases, at least for "child" Opportunities e.g. Slots and ScheduledSessions, the full opportunity data is stored once in both caches.

For this issue, investigate whether it's possible to consolidate these such that we don't store two copies of each child Opportunity. One possible outcome is that there's a good reason for it to stay like it is (although I expect we could at least change one of the storage methods to use an ID to refer to the other). Try to get a good understanding of why it is the way it is before changing it. There might be some wisdom in there, which may still be true or now be outdated.

Issue #667, opened 2024-03-25
Elaborate on or implement all TODOs in Test Suite


lukehesluke:
Ensure they are dealt with or adequately documented

lukehesluke:
Output issues:

* https://github.com/openactive/openactive-test-suite/issues/666

PR #670, opened 2024-03-25
chore(broker): Categorise opportunity caches

lukehesluke:
Closes https://github.com/openactive/openactive-test-suite/issues/668

(also, this includes a small amount of work to elaborate TODOs: https://github.com/openactive/openactive-test-suite/issues/667)

The purpose is to make it clearer what all the various ES Maps are in the Broker state and why there are so many data structures holding data that relates to the caching of opportunities.

This work also ellucidated that the opportunity data is duplicated. An issue has been made for that here: https://github.com/openactive/openactive-test-suite/issues/669

### DIFF
--- a/packages/openactive-broker-microservice/src/state.js
+++ b/packages/openactive-broker-microservice/src/state.js
@@ -2,7 +2,7 @@ const { OpenActiveTestAuthKeyManager } = require('@openactive/openactive-openid-
 const config = require('config');
 const { TwoPhaseListeners: Listeners } = require('./twoPhaseListeners/twoPhaseListeners');
 const PauseResume = require('./util/pause-resume');
-const { OpportunityIdCache } = require('./util/opportunity-id-cache');
+const { CriteriaOrientedOpportunityIdCache } = require('./util/criteria-oriented-opportunity-id-cache');
 const { log } = require('./util/log');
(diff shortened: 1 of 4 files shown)


packages/openactive-broker-microservice/src/state.js
```diff
@@ -101,78 +101,113 @@ const state = {
...
+  /* TODO investigate consolidation of opportunityItemRowCache and
+  opportunityCache to reduce memory usage & simplify code:
+  https://github.com/openactive/openactive-test-suite/issues/669 */
   /**
    * A criteria-oriented cache for opportunity data. Used to get criteria-matching
    * opportunities for tests.
    */
-  opportunityIdCache: OpportunityIdCache.create(),
-  // nSQL joins appear to be slow, even with indexes. This is an optimisation pending further investigation
+  criteriaOrientedOpportunityIdCache: CriteriaOrientedOpportunityIdCache.create(),
```
civsiv, line 111: Maybe a comment linking to the schema of this cache? Just thinking about if someone hovers over `criteriaOrientedOpportunityIdCache` it's not clear what the schema is (which is not the case hovering over other maps)

lukehesluke: I'll leave this for now because CMD-clicking (in VSCode) will get you there

PR #696, opened 2024-10-15
feat: Separate out past-opportunities feature

lukehesluke:
So that it may be opted out of if no past data is available (esp. if using random mode)

This was prompted by Bookteq (and I've seen it for other users too), whose random modes are not capable of producing opportunities in the past.

There is a `not-implemented` test to ensure that there are no such opportunities if this feature is turned off.

----

Even still, it is advisable to keep this feature on if at all possible. Even if a booking system cannot generate opportunities in the past in test suite conditions, it may be able to have opportunities expire but remain in the feed outside of test suite conditions. This is a compromise to help accelerate adoption but it may be worth considering if it should be kept as required


nickevansuk (approved):
Looks great @lukehesluke !

Issue #698, opened 2024-10-15
Allow Broker to ignore irrelevant updates when listening for Order/Opportunities


lukehesluke:
## Example Scenario

_this is a scenario which we have seen with an organization attempting to pass Test Suite on their booking system_

1. Run Seller Requested Cancellation on an Order with multiple OrderItems. The test has invoked the `SellerRequestedCancellationSimulateAction` and is waiting for a change to the Orders RPDE feed (using Broker's `/order-listeners/:type/:bookingPartnerIdentifier/:uuid` endpoint)
2. Booking System's processes are set up in such a way that it cancels each OrderItem one at a time
3. Boker sees the first change to the Orders RPDE feed in which just one of the OrderItems is designated as cancelled and passes this back to Test Suite
4. Test Suite fails validation as only one of the OrderItems is cancelled and it was expecting all to be.

## A possible solution

Suppose Broker was given instructions on what kind of updates it should expect to see, and so what it can ignore. Then, in the above scenario, Broker would ignore the first Orders RPDE feed update, because it's expecting to see _all_ OrderItems be cancelled. It would keep ignoring them until it finds an update which sets all to cancelled. Then, the test would pass

## Potential issues

I would not want to implement something like this until RPDE tracing (https://github.com/openactive/openactive-test-suite/issues/665) has been implemented.

If we implemented this without RPDE tracing then this is what would happen when a user has a test failure (using again a Seller Requested Cancellation test example):

1. Test Suite invokes `SellerRequestedCancellationSimulateAction` and asks Broker to listen for the Orders RPDE feed change, which should have OrderItems set to seller-cancelled.
2. Broker sees something in the Orders RPDE feed which is incorrect and ignores it
3. Then, Broker times out
4. The user looks at the Test Suite results and just sees `ESOCKETTIMEDOUT` and has no clue what they did wrong

When we have RPDE tracing, we could have the Test Suite results include info about each update that was found and why it was ignored if it was.

lukehesluke:
This PR: https://github.com/openactive/openactive-test-suite/pull/699 implements the **Possible Solution**, but only for Seller Requested Cancellation. Further PRs would be needed to also enable it for customer cancellation, etc.

PR #700, opened 2024-11-01
feat: Use ListenerItemExpectations for customer cancel

lukehesluke:
Implements #698 for customer-requested cancellations

## QA

I tested this against a booking system which is not yet stable, but was able to gain confidence that the feature works as expected.

These screenshots demonstrate some of this. They are a side-by-side comparison of the results for the `book-and-cancel` test. The LHS includes the changes and the RHS does not:

**Single mode**: 
Screenshot 2024-11-01 at 11 31 04

Here, the booking system failed to produce the CustomerCancelled order feed change in the LHS test, but this lead to it failing at the order feed download step, as it could never be found. Whereas, in the RHS test, Test Suite passes the order feed download step but then later fails when asserting about orderItemStatuses

**Multiple mode**: 
Screenshot 2024-11-01 at 11 32 04

Here, the booking system succeeded in producing the CustomerCancelled order feed change, and so the LHS test passed (except an unrelated validation issue), whereas the RHS did not, because Broker latched onto an irrelevant order feed update (this booking system is one that produces such updates)


lukehesluke:
[Bookteq/Vivify: Expand ListenerItemExpectations to customer-requested cancellations](https://trello.com/c/0FsaeAKe/5234-bookteq-vivify-expand-listeneritemexpectations-to-customer-requested-cancellations)

nickevansuk (approved):
Nice!

PR #714, opened 2025-04-02
feat: add Accept header for integration tests

civsiv:


packages/openactive-integration-tests/test/helpers/request-helper.js
```diff
@@ -169,6 +169,7 @@ class RequestHelper {
 
   createHeaders() {
     return {
+      'Accept': 'application/vnd.openactive.booking+json; version=1',
```
civsiv, line 172: If I've understood content negotiation properly, this is needed for all requests to Bookings Systems

packages/openactive-integration-tests/test/helpers/request-helper.js
```diff
@@ -169,6 +169,7 @@ class RequestHelper {
...
       'Content-Type': 'application/vnd.openactive.booking+json; version=1',
```
civsiv, line 173: This is needed to say what the content type of the request is

lukehesluke (changes requested):
@civsiv it looks like the Broker has two places where it sets the wrong Accept headers also. It sets `Accept: 'application/json, application/vnd.openactive.booking+json; version=1',`, which has an extraneous `application/json`.

It is possible there's some legacy reason for this

lukehesluke:
Also, packages/openactive-integration-tests/test/global-setup.js's `deleteTestDataset(..)` needs Accept header (https://openactive.io/test-interface/#delete-test-interfacedatasetstestdatasetidentifier)

Issue #715, opened 2025-04-02
Broker: Support large feeds


lukehesluke:
If you run Broker against a large feed with, say, 100,000s of items, it will eventually crash due to running out of memory.

This currently forbids booking systems which represent large leisure operators to use Test Suite in random mode (without lots of effort on their part to create special pruned test environments).

It should be possible for Broker to cache its items in such a way that it can support feeds of arbitrary size. For example, by caching items to disk, or to an external storage system.

PR #716, opened 2025-04-08
chore(broker): Create PersistentStore

lukehesluke:
PersistentStore contains opportunity caches. This refactoring prepares Broker Microservice for bringing its opportunity caches (which scale with feed size and so can become huge - see https://github.com/openactive/openactive-test-suite/issues/715) out of memory, so that it can support larger feeds.

# Expanded Rationale

I am working on https://github.com/openactive/openactive-test-suite/issues/715.

Broker's approach to state is quite messy at the moment. State is updated and read from all over the place (and somtimes in quite convoluted ways). My thinking is that in order to support moving opportunity caches to something like SQLite, I have to first create an interface for that data. Then, the implementation of that interface can change from in-memory objects to SQLite (or something else).

So this PR respresents step 1: creating this interface.


packages/openactive-broker-microservice/src/core.js
```diff
@@ -904,7 +834,7 @@ async function storeChildOpportunityItem(item) {
  * Merge a child- and parent- opportunity and then save them to the
  * criteria-oriented cache and notify any listeners.
  *
- * @param {import('./models/core').OpportunityItemRow} row
+ * @param {Readonly<import('./models/core').OpportunityItemRow>} row
```
lukehesluke, line 837: Make input data readonly to be sure that they aren't secretly doing mutations (which would not work as expected if data is being stored in something outside of JS memory)

packages/openactive-broker-microservice/src/util/get-opportunity-by-id-from-cache.js
```diff
@@ -5,25 +5,24 @@ const { jsonLdHasReferencedParent, getMergedJsonLdContext } = require('./jsonld-
...
   if (!jsonLdHasReferencedParent(opportunity)) {
     return opportunity;
   }
-  const superEvent = state.opportunityCache.parentMap.get(/** @type {string} */(opportunity.superEvent));
-  const facilityUse = state.opportunityCache.parentMap.get(/** @type {string} */(opportunity.facilityUse));
+  const superEvent = state.persistentStore.getOpportunityCacheParentItem(/** @type {string} */(opportunity.superEvent));
+  const facilityUse = state.persistentStore.getOpportunityCacheParentItem(/** @type {string} */(opportunity.facilityUse));
   if (superEvent || facilityUse) {
     const mergedContexts = getMergedJsonLdContext(opportunity, superEvent, facilityUse);
-    delete opportunity['@context'];
```
lukehesluke, line 23: Sneaky (unintentioned?) mutation

packages/openactive-broker-microservice/src/util/persistent-store.js
```diff
@@ -0,0 +1,488 @@
...
+ * misinterpretation that this will update the data itself (as if it was
+ * in-memory).
+ */
+class PersistentStore {
+  constructor() {
+    /**
+     * A criteria-oriented cache for opportunity data. Used to get criteria-matching
+     * opportunities for tests.
+     */
+    this._criteriaOrientedOpportunityIdCache = CriteriaOrientedOpportunityIdCache.create();
```
lukehesluke, line 28: All of these values and their docs copied over from state.js (and with the `_` prefix added - making them super private 🕵️ )

packages/openactive-broker-microservice/tsconfig.json
```diff
@@ -6,7 +6,8 @@
     "downlevelIteration": true,
     "target": "ES2019",
     "moduleResolution": "node",
-    "resolveJsonModule": true
+    "resolveJsonModule": true,
+    "noImplicitAny": true
```
lukehesluke, line 10: I moved lots of state from state.js to PersistentStore. After making this change, TS didn't show me all the places that needed to be fixed right away because its default functionality (without this flag) is to treat a literal object as being `{ <...allDefinedFields>, [k: string]: any }`. So if you consider `state.asdkfjlasjdflkasjfd`, TS will happily treat it as an `any` 🙃.

I think this is an improvement on the codebase too (it's amazing what TS allows without this), though unfortunately for the reviewer of this PR, it does mean that a lot of the change is just improving various types as was required when turning this on

civsiv: I could not agree more, v type safe!

civsiv (approved):

PR #718, opened 2025-04-10
chore(broker): Consolidate opportunity caches

### LINKED ISSUE #669: Broker: Consolidate opportunityItemRowCache and opportunityCache to reduce memory usage & simplify code

lukehesluke:
Broker has a large number of caching strategies which have been developed at different times with different aims. Much of it can be consolidated, which reduces memory usage, and simplifies the code.

This will also make it easier to move caching into a database (#715)

Closes #669


packages/openactive-broker-microservice/app.js
```diff
@@ -100,7 +100,7 @@ state.incompleteFeeds.markFeedHarvestStarted('DatasetSite');
 const server = http.createServer(app);
 server.on('error', onHttpServerError);
 
-app.listen(PORT, () => {
+app.listen(PORT, async () => {
```
lukehesluke, line 103: I disentangled this PR from ongoing work on moving the cache into the database. Unfortunately it would be a lot of effort to revert moving all cache actions to async. Hopefully this is okay to include as it doesn't functionally change anything 🙏

packages/openactive-broker-microservice/src/util/persistent-store.js
```diff
@@ -18,11 +18,6 @@ const { mapToObjectSummary } = require('./map-to-object-summary');
  * does not attempt to mutate it, as this likely means that it is under the
  * misinterpretation that this will update the data itself (as if it was
  * in-memory).
- *
- * We use multiple strategies to cache opportunity data for different use cases.
```
lukehesluke, line 22: Fixed

civsiv (approved):

